PR #1, opened 2022-12-01
First draft of application schema

domire8:
Here is a first and incomplete draft of the application schema. I propose to do this incrementally and merge the PRs to a branch `feature/schema`.

Dockerfile, serve-html.sh and validate.sh are copied from the component sdk.

Then, the schema currently supports the following:
- Top level properties `on_load`, `hardware_interfaces` and `components`
- `on_load` supports `load`, `load_controller` and  `switch_controller`. each of these items can either be an array or just one object
- `hardware_interfaces` supports array or one item of a named hardware interface with one or several controllers that each have a name, plugin, and parameters
- `components` supports an array or one item of a named component with a class name and one or several parameters

What's missing currently (besides other things) are the events that can be added to a component. This will be an object with many possible properties and sub-objects, which is why I could do that in a next PR.

@buschbapti


applications/schema/schema/application.schema.json
```diff
@@ -0,0 +1,61 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "YAML Application Description",
+  "description": "A full description of a YAML application AICA framework.",
+  "type": "object",
+  "additionalProperties": false,
```
eeberhard, line 6: I would remove this, especially since we are missing some fields like events. Even if we have a complete description, I don't see a reason to prohibit metadata or other fields.

applications/schema/schema/components.schema.json
```diff
@@ -0,0 +1,37 @@
...
+    "^[a-zA-Z][a-zA-Z0-9_]*[a-zA-Z0-9]$": {
+      "title": "Component Name",
+      "description": "The name of the component",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "component": {
+          "title": "Component",
+          "description": "The name of the component class",
+          "type": "string"
```
eeberhard, line 17: do you think we should move away from the option to specify "package" and "class" separately? In the component description we support both kinds of registration, but it could be simpler to just require the "new" (1 year old) reduced registration syntax.

eeberhard: To be specific and match the component SDK, I would change the description field to 
"The fully qualified class name used to register the component."

eeberhard (commented):
This is a good start. We can keep it simple to begin and add more things in future PRs on the side branch as you suggest.

applications/schema/schema/application.schema.json
domire8, line 6:
```suggestion
  "type": "object",
```

applications/schema/schema/components.schema.json
domire8: For your first question, yes I don't see the advantage of keeping the old kind too, its and additional (and even a bit duplicated) line.

applications/schema/schema/components.schema.json
```diff
@@ -0,0 +1,37 @@
...
+  "patternProperties": {
+    "^[a-zA-Z][a-zA-Z0-9_]*[a-zA-Z0-9]$": {
+      "title": "Component Name",
+      "description": "The name of the component",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "component": {
+          "title": "Component",
+          "description": "The name of the component class",
```
domire8, line 16:
```suggestion
          "description": "The fully qualified class name used to register the component",
```

applications/schema/schema/load.schema.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
```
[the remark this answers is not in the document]
domire8: Yes indeed, maybe a subfolder would be nice. I could imagine that one single file would be a looot of lines

applications/schema/schema/parameter.schema.json
```diff
@@ -0,0 +1,76 @@
+{
```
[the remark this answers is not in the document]
domire8: Yeah sorry this shouldn't be here. I'm not sure if we need to go down all these levels and really check all the parameter values too. If yes, thats one of the last steps probably

applications/schema/schema/application.schema.json
```diff
@@ -0,0 +1,61 @@
...
+        {
+          "type": "array",
+          "items": {
+            "$ref": "hardware_interfaces.schema.json"
+          }
+        },
+        {
+          "$ref": "hardware_interfaces.schema.json"
+        }
+      ]
```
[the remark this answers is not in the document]
domire8: Well because if you just have
```
hardware_interfaces:
  papa:
    urdf: /home/ros2/examples/urdf/franka_panda.urdf
    rate: 500
    controllers:
      broadcaster:
        plugin: joint_state_broadcaster/JointStateBroadcaster
      twist_controller:
        plugin: compliant_twist_controller/CompliantTwistController
        parameters:
          linear_principle_damping: 50.0
          linear_orthogonal_damping: 50.0
          angular_stiffness: 3.0
          angular_damping: 3.0
```
this will be parsed as an object and then you get a warning that you have an object instead of an array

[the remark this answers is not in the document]
domire8: Let me explain a bit more, maybe you are aware of this, maybe not.
```
hardware_interfaces:
  papa:
    urdf: 1
  quebec:
    urdf: 2
```
In the case above, the schema interprets everything below `hardware_interfaces` as an object with this syntax, whereas
```
hardware_interfaces:
  - papa:
      urdf: 1
  - quebec:
      urdf: 2
```
is interpreted as array of objects. So if we were to remove all those `oneOf`s (there are several, for the same reason) we would have to keep the object and discard the array of object, because as mentioned above, type array is not possible for our current apps, only if we would enforce the `- ` which is a bit weird for yaml files.

Therefore, I decided to support both variants with all these `oneOf` cases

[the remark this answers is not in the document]
eeberhard: Sorry for the late followup to this review. I guess I got it the wrong way around in my original comment, but the main point is that we shouldn't support both if one way is "more correct". The schema should just allow an object of type `"$ref": "<subobject>.schema.json"` and not the array of objects. I would say that if the YAML interface can actually parse an array of interface objects with the `- papa, - quebec` notation, then it is an unintentional side-effect and not officially supported.

domire8:
This should then be good to go as a start or did I miss something urgent from your review?

PR #2, opened 2022-12-18
Extend application schema with events

eeberhard:
* Define event schemas in subfolder

* Reuse common identifiers for component and hardware names where possible

* Add log level and mapping fields to components

* List common predicates in component events field

* Generally refine titles and descriptions

* Add type constraints for parameter fields

* Refine regex for component and parameter names to allow single-character names

---

I had some time on the train today to / from a family event. Sorry for the large diff and partial rewrites of what you already had done, but I felt inclined to advance this. This schema will be useful for onboarding new users to the YAML syntax, but will also let the "set_application" API call validate the application in a more sensible way.

I also have a revised README from the state engine which should live here, but I'll add that in a separate PR. After this it's mainly just the conditions that are missing, which will require some recursive references to validate but should be feasible. I'll probably do that over the holidays to be honest.


applications/schema/schema/events/switch_controllers.schema.json
```diff
@@ -0,0 +1,66 @@
...
+              }
+            },
+            {
+              "$ref": "identifiers/controller.schema.json#/properties/controller"
+            }
+          ]
+        }
+      },
+      "required": [
+        "interface"
```
domire8, line 62: Here, would it be possible to do something like
```
allOf
  required [interface]
  anyOf
     required [ start]
     required [stop]
```
to make sure at least start or stop is there?

domire8 (commented):
A few comments, but it looks very clean! Nice

applications/schema/schema/components.schema.json
```diff
@@ -1,25 +1,132 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "title": "Components",
-  "description": "components",
+  "description": "A description of components used in the application",
   "type": "object",
   "additionalProperties": false,
   "patternProperties": {
-    "^[a-zA-Z][a-zA-Z0-9_]*[a-zA-Z0-9]$": {
+    "(^[a-zA-Z][a-zA-Z0-9_]*[a-zA-Z0-9]$)|(^[a-zA-Z]$)": {
```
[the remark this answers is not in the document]
eeberhard: It's technically possible and permitted to have single-character components (`a`, `b`...) which would have failed validation before. This is more apparent in the case of parameters where could easily just have `x`, `y`.

applications/schema/schema/components.schema.json
```diff
@@ -1,25 +1,132 @@
...
+                "boolean",
+                "number",
+                "string",
+                "array"
+              ],
+              "items": {
+                "type": [
+                  "boolean",
+                  "number",
+                  "string"
```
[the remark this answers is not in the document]
eeberhard: The list of types gives the permitted types, which can include array. Then the `items` qualifier only applies to the array type for permitted types within an array. This way we can permit all values and arrays of values, but not mixed-type arrays, dicts, or nested arrays.

So, in the end we allow `bool`, `double`, `string` or `bool[]`, `double[]`, `string[]`. In this case `number` covers both `integer` and `double`.

applications/schema/schema/events/switch_controllers.schema.json
eeberhard: Yes I realise I took that out from your previous version. I will check again but I think it was causing problems with the validation ("the schema validates to multiple variants").

eeberhard: Ok you are correct, the `allOf` container works and solves the multiple variant problem 👍

domire8 (approved):

PR #3, opened 2022-12-21
Rename application service endpoints

eeberhard:
Upcoming changes to the application API endpoints are intended to increase the clarity of behaviours. The following endpoints are affected: 

- `stop_application` has been renamed to `pause_application`
- `reset_application` has been renamed to `stop_application`

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -13,14 +13,12 @@ from aica_api.client import AICA
 aica = AICA()
 
 aica.set_application('my_application.yaml')
-aica.init_application()
 aica.start_application()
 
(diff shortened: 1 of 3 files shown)


domire8 (approved):
This looks good, can I test it?

eeberhard:
> This looks good, how can I test it?

To test it locally, I use `pipenv install ./python` in the root directory of the repo to create a Pipfile and virtual environment. Then you can do `pipenv run python` and test it like that. 

I would only push the project to PyPi once it's merged (which can be made into a CI action eventually)

PR #4, opened 2022-12-21
Add YAML application syntax documentation

eeberhard:
Most of this was taken from the existing README in the state engine. I reformatted and extended it with descriptions of the hardware interfaces and conditions.

### DIFF
--- a/applications/README.md
+++ b/applications/README.md
@@ -1,0 +1,488 @@
+# AICA Applications
+
+An AICA application is interconnected graph of **components** and **hardware interfaces**. Components process data in a
+periodic step function and transfer data as signals to other components and hardware interfaces. Hardware interfaces
+are drivers that connect controllers to robots. The application state can be dynamically changed through **events**,
+which can change parameters, load or unload components, trigger specific service actions and more.
+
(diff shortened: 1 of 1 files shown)


applications/README.md
```diff
@@ -0,0 +1,488 @@
...
+
+
+## On Start
+
+The `on_start` keyword is reserved as a special event trigger when the application is launched.
+List the [events](#events) to trigger on startup (for example, to load components).
+
+```yaml
+on_start:
+  load:
```
domire8, line 61: You know, now that I read this again, and we are renaming state engine methods, wouldn't it be more correct to have `load_component` and `unload_component`. To a user it might seem unclear why components are somehow more important than controllers and hardware

domire8: On second thought, I think its okay, because otherwise `set`, `service`, and `lifecycle` would almost need a more verbose name too

domire8 (approved):
Thanks for making the effort to polish this README! The users and I appreciate it.

applications/README.md
eeberhard: It's a fair point but I think component-centric events are the default so it's nice to have the brevity. As long as it's well documented I think it's ok like this.

PR #5, opened 2022-12-23
Extend application schema with conditions

eeberhard:
To wrap up the schema, here are conditions. It was actually not as bad as I thought to incorporate the recursive logic. A quick summary:

Top-level must have `events` and either `condition_object` or `predicate_object` fields. The predicate object is just an associated component and predicate. The condition object has fields [all, any, one_of, not]. The first three refer to `condition_array` - that is, a list of items which must be either another condition object, predicate object, or bool (for hard-coded conditions). The `not` doesn't take an array of items, but instead just one of condition object, predicate object, or bool.

Here are some valid examples that are validated correctly:

```yaml
# valid conditions
conditions:
  test_single_predicate:
    component: foo
    predicate: foo
    events:
      load: foo
  test_single_not:
    not:
      component: foo
      predicate: foo
    events:
      load: foo
  test_nested_condition:
    not:
      all:
        - { component: foo, predicate: is_foo }
        - { component: test_component_01, predicate: is_foo }
        - any:
            - { component: test_component_02, predicate: is_foo }
            - { component: test_component_03, predicate: is_foo }
            - one_of:
                - { component: unregistered_component, predicate: is_foo }
                - false
                - true
    events:
      load: foo
```

Here are invalid conditions that are correctly identified
```yaml
# invalid conditions
conditions:
  test_no_events:
    component: none
    predicate: none
  test_no_condition:
    events:
      load: foo
  test_multiple_fields:
    all:
      - true
      - false
    any:
      - true
      - false
    events:
      load: foo
  test_invalid_subobject:
    all:
      - { not: a, valid: object }
    events:
      load: foo
  test_listed_not:
    not:
      - true
      - false
    events:
      load: foo
```

@buschbapti

### DIFF
--- a/applications/schema/schema/application.schema.json
+++ b/applications/schema/schema/application.schema.json
@@ -33,6 +33,9 @@
     "hardware_interfaces": {
       "$ref": "hardware_interfaces.schema.json"
     },
+    "conditions": {
+      "$ref": "conditions.schema.json"
+    },
     "components": {
       "$ref": "components.schema.json"
     }
--- a/applications/schema/schema/components.schema.json
+++ b/applications/schema/schema/components.schema.json
@@ -74,18 +74,18 @@
           }
         },
         "events": {
-          "title": "Predicate events",
+          "title": "Predicate Events",
           "description": "Component predicates that trigger events",
           "type": "object",
           "additionalProperties": false,
           "properties": {
             "on_load": {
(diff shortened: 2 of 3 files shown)


domire8 (approved):
That's very neat! Quite cool how you can validate all these cases with just a few defs

Issue #10, opened 2023-02-23
Extend YAML Application schema to add more abstractions


eeberhard:
The schema is still very close to the implementation. We can easily add some more abstractions to make the syntax more clear. 

A main improvement is for signals, which have topics coded by specially named parameters. This could be improved as follows:

```yaml
components:
  foo_old:
    component: foo::Foo
    parameters:
      x: 1
      a_topic: /foo
      b_topic: /bar

  foo_new:
    component: foo::Foo
    parameters:
      x: 1
    inputs:
      a: /foo
    outputs:
      b: /bar
```

An additional field may be to give the component a nicer display name `name: "Signal Processing"`, rather than the unique snake case name (`signal_processing`).

More abstractions in the schema will bring the YAML syntax closer to the frontend representation of applications, rather than the backend interpretation.

After making this change in the application schema, the YAML application interface in the state engine should be updated accordingly.

PR #11, opened 2023-03-09
Extend component YAML schema with more abstractions

eeberhard:
* Add a display_name property that allows the component to have a nicer, human-readable name (i.e. one that doesn't need to be lower_snake_case)

* Add signals properties (inputs / outputs) that would replace occurences of `foo_topic` parameters

## Description

- #10 

This PR solves the issue by adding the suggested elements to the YAML schema. It is a non-breaking change as the new fields are just optional abstractions.

## Review guidelines

## Related issues

No issue has been created yet, but as a follow-up action I would implement the signal abstraction in the state engine YAML parsing.

### DIFF
--- a/applications/schema/schema/components.schema.json
+++ b/applications/schema/schema/components.schema.json
@@ -23,6 +23,11 @@
             "fatal"
           ]
         },
+        "display_name": {
+          "title": "Component Display Name",
+          "description": "The human-readable name to display on the component",
+          "type": "string"
+        },
         "component": {
           "title": "Component Registration",
           "description": "The fully qualified class name used to register the component",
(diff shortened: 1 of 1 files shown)


domire8 (approved):
Good that it is optional for now, do you plan on making it non-optional in the near future?

eeberhard:
> Good that it is optional for now, do you plan on making it non-optional in the near future?

It would be very difficult to enforce. It would have to perform YAML syntax validation, then look up the matching component description to see what inputs / outputs there should be, and then change the outcome of the validation accordingly. Only the display name could be made required, but there we always have the fallback option of using the YAML component name (or the display name from the component description).

The only thing I can think is to prohibit parameters that end in `_topic`, but I wouldn't go that far.

PR #19, opened 2023-07-13
Update Python API client to use new v1 endpoints

jjenscodee:
## Description

- #14 

This PR solves the issue by changing the endpoint to v1.

## Review guidelines


eeberhard (approved):
Looks good to me! As discussed, we keep the same "simple" functions in the python client for now and only change the endpoints they use. In future we can decide if the python client should also be differently structured (something like `AICA.applications.components.load()`) but generally this pattern is quite nice to use.

Issue #15, opened 2023-06-29
Automatically convert component description and application schemas to TypeScript declarations


eeberhard:
When a change to the component SDK / description schema or api / application schema is pushed to develop, it should trigger a workflow to compile a `.d.ts` TypeScript declaration file (e.g. with `json-schema-to-ts`) and host / push the files to the frontend.

It should also run a formatting pass on the generated output with `prettier` so that it matches the project settings.

PR #21, opened 2023-07-14
Automatically convert component description and application schemas to TypeScript declarations

jjenscodee:
## Description

- #15 

This PR solves the issue by creating a workflow that generates TypeScript declarations and pushes the files to the frontend repository when schemas are updated.

## Review guidelines


.github/workflows/generate-ts-declaration.yml
```diff
@@ -0,0 +1,69 @@
...
+          token: ${{ secrets.FRONTEND_REPOSITORY_WRITE }}
+      - name: Create a new branch in frontend repository
+        run: |
+          cd frontend
+          ls
+          git config --global user.name github-actions[bot]
+          git config --global user.email github-actions[bot]@users.noreply.github.com
+
+          git checkout -b ci/update-schemas
+          git push -u -f origin ci/update-schemas
```
eeberhard, line 41: Since you already checkout the other repo and push a branch here, it seems it would be more efficient to just add the new generated files in this step. That way you can skip the two `write-file` steps afterwards.

You would first have to do both `generate-typescript-declaration` actions, move the results manually, then just add and commit.

```suggestion
          mv schemas/applications/schema/application.d.ts frontend/src/utils/schemas/
          mv schemas/component-descriptions/schema/dynamic-component-description.d.ts frontend/src/utils/schemas/

          cd frontend
          git config --global user.name github-actions[bot]
          git config --global user.email github-actions[bot]@users.noreply.github.com

          git checkout -b ci/update-schemas
          git add src/utils/schemas
          git commit -m "ci: update typescript declaration files from new schemas"
          git push -u -f origin ci/update-schemas
```

It's nice to re-use the write-file action where possible but for this particular case it's not well suited because we require a new feature branch on the remote repo first.

eeberhard (commented):
Seems to do the job as desired, well done. Let me know what you think about my comment to reduce the number of steps.

eeberhard (approved):
Nice work!

.github/workflows/generate-ts-declaration.yml
```diff
@@ -0,0 +1,48 @@
...
+          schema: component.schema.json
+          ts-declaration: dynamic-component-description.d.ts
+
+      - name: Checkout frontend repository
+        uses: actions/checkout@v3
+        with:
+          repository: aica-technology/frontend
+          ref: develop
+          path: frontend
+          token: ${{ secrets.FRONTEND_REPOSITORY_WRITE }}
```
eeberhard, line 36: According to https://github.com/peter-evans/create-pull-request/issues/48#issuecomment-584209254, if we use SSH deploy keys, then the push can be used to trigger GitHub actions.

```suggestion
          ssh-key: ${{ secrets.FRONTEND_REPOSITORY_WRITE_SSH }}
```

eeberhard (approved):
I think this is good to go 👍

LouisBrunner:
Just as a note on this, I find it very odd that this repo would _push_ changes to the Front-End repo.

This definition is used by our software and our users so we should probably version it and host it so that we can refer to different versions easily. Then other components (e.g. Front-End) would pull it based on which version they want to build.

We could still have a PR system (but also the system of creating a branch which then creates a PR seems very disconnected) for changes but right now it's very difficult to keep track which version the Front-End is using.

eeberhard:
Fair point, thanks for the insight! I certainly agree that tagging fixed versions of the schemas in this repo is important come release and moving forwards. I was considering that, once we merge to main and tag a first `1.0`, then this workflow could write the `.d.ts` files to FE with an explicit version in the filename. That way it's clear from the file and import usage in the stack what version of the schema is being used, and potentially some conversion utilities could add some sort of backwards compatibility then.

But if you think pushing from this branch to the other is a weird pattern and rather have a FE process to fetch tagged versions, that's fine with me.

LouisBrunner:
I guess I imagine having a central place to store those reference schemas and having other apps pull them makes more sense than a pushing patterns. We might want other components (e.g. a Python SDK or our backend) to pull them for validation purpose and a pushing pattern from one repository to many seems more difficult to maintain than a pulling one (especially as it means all the quirks of those repos are stored here instead of their respective repos).

PR #20, opened 2023-07-13
Controller description schema and reusable parameter and signal schemas

eeberhard:
## Description

- https://github.com/aica-technology/modulo-controllers/issues/91

This PR solves the issue by adding a simple description schema for controllers. For now it just covers signals and parameters, though controller services and predicates are a future idea. 

To avoid code duplication of elements in the component schema, I moved the common parts to separate schemas (parameters and signals) and reference them through the raw GitHub link. For now it's referencing the feature branch so that you can test it out (serve-html.sh) but there is a TODO to change those before merging.

## Supporting information
In case you don't want to run the `serve-html.sh` script, here's the saved html output for the controller schema (compressed as ZIP to make GitHub happy)
[AICA Controller Description.html.zip](https://github.com/aica-technology/api/files/12044453/AICA.Controller.Description.html.zip)

## Review guidelines

Estimated Time of Review: 10 minutes

TODO before merging:
- [x] change paths in ids and refs from `aica-technology/api/feature/controllers-schema` to `aica-technology/api/develop`

PR #22, opened 2023-07-15
Controller inheritance and internal parameters

eeberhard:
## Description

- https://github.com/aica-technology/modulo-controllers/issues/91

Following #20, I also wanted to have the ability to mark inheritance for controllers. However, many of the base parameters in the modulo controller interface are "optional" or only used internally. Derived controllers often will not need to set this, and in general the user-facing parameters should be reduced where possible. The new `internal` property for parameters allows us to mark this behavior and thereby hide some parameters from the user (unless they tick some box to show all parameters for example).

I did it in a separate step because I wanted to add a field to the parameters descriptions, but not introduce too many changes in a single PR.

Up to debate whether `internal` is a good name or something like `advanced` would be preferred.

## Review guidelines

### DIFF
--- a/schemas/component-descriptions/README.md
+++ b/schemas/component-descriptions/README.md
@@ -209,9 +209,14 @@ default state and must be set by the user for the component to function, the `de
 description should be set to `null`. This is distinct from a valid default empty parameter state, which can be
 expressed with the empty string `""`.
 
-If a parameter is dynamically reconfigurable, it implies that the parameter can be changed after the component
-is already configured to influence the run-time behaviour.
+If a parameter is dynamically reconfigurable, as described by the `dynamic` property, it implies that the parameter can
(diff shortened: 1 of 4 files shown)


domire8:
I'm not sure what you mean by "optional" or "internal" parameters. I would personally understand "optional" as parameters that have a good default value, but what are examples of "internal" parameters?

eeberhard:
The parameters `hardware_name`, `urdf_path`, `urdf_content`, `joints`, `sort_joints`, are all "internal" in the sense that they are either handled by the state engine automatically, or in the case of sort_joints generally never have to be set false because it's about how the JointState object is created.

For me it's not just about "optional" parameters with sensible defaults but rather for parameters that are quite technical to how the stack works, so "advanced" or "internal".

The main incentive is to hide such parameters in the UI because otherwise the controllers get very crowded with parameters that are never touched.

Issue #23, opened 2023-07-17
Add component position and buttons to application schema


eeberhard:
The features https://github.com/aica-technology/frontend/pull/212 and https://github.com/aica-technology/frontend/pull/221 have new expectations of the application schema, namely to support XY positioning of components and interactive buttons. The changes were made to the local `application.d.ts` file.

However, these changes should be made upstream in the JSON schema itself, so that the workflow to auto-generate new typescript declarations does not break the existing features.

PR #27, opened 2023-07-18
Add component position and buttons to application schema

jjenscodee:
## Description

- #23 

This PR solves the issue by adding component position and buttons to application schema.

## Review guidelines

### DIFF
--- a/schemas/applications/schema/application.schema.json
+++ b/schemas/applications/schema/application.schema.json
@@ -38,6 +38,12 @@
     },
     "components": {
       "$ref": "components.schema.json"
+    },
+    "buttons": {
+      "additionalProperties": {
+        "$ref": "buttons.schema.json"
+      },
+      "type": "object"
     }
   },
   "required": [
--- a/schemas/applications/schema/application.schema.json
+++ b/schemas/applications/schema/application.schema.json
@@ -40,10 +40,7 @@
       "$ref": "components.schema.json"
     },
     "buttons": {
-      "additionalProperties": {
-        "$ref": "buttons.schema.json"
-      },
-      "type": "object"
+      "$ref": "buttons.schema.json"
     }
   },
   "required": [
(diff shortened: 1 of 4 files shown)


eeberhard:
A good start but a few quirks and inconsistencies, especially compared to the rest of the schema. Maybe that's because I've been writing JSON schemas the wrong way until now; but I would prefer to keep things consistent in this repository. 

For one, you use `additionalProperties` in interesting ways, sometimes defining it as empty `{}` but other times defining actual properties inside. Normally, `additionalProperties` is a true / false flag that indicates if an object can contain additional properties or not, with no. You also chose interesting names (`ButtonName1`) when something more simple / readable could be preferred. Finally, the ordering of schema files.

I hope you don't mind, I will push a commit with suggested changes. See the commit message for details on the changes.
A consequence of my suggested changes will be that you will need to revise the actual implementation in the frontend to respect the new interface, but I think that will be a manageable change. I would prefer to have a clean and readable schema + application yaml, and revise the frontend implementation from there, then try to match the schema to the existing type names in the implementation.

Consequences for the frontend implementation, specifically:
- rename interface `ButtonName1` to `Buttons`
- rename property `events` to `on_click`
- because of reusing the `load` and `unload` event subschemas, technically, `on_click` events would now support a string _or_ an array, as is the case with component events. Example below:
```yaml
buttons:
  button1:
    position:
      x: 0
      y: 0
    on_click:
      load: foo  # if there is only one event, it can be a string, not an array of strings
  button2:
    position:
      x: 100
      y: 0
    on_click:
      load:      # if there are multiple events, it's an array of strings
        - foo
        - bar
```
This last point means you should check the type first (`if (typeof button.on_click.load === "string")`) and only do a `.map` operation in case of array. 

Anyway hope that's clear, let me know if you have any questions or concerns!

jjenscodee:
Thank you for the changes and advice. It's my first time writing the schema json, I'll be more careful to the naming and consistency next time!

eeberhard (approved):
The JSON schema syntax is not super nice / readable in the best case, and there's not really a right or wrong way to do it. To keep things moving forward, I thought it would be more efficient to implement the suggestions directly rather than try to explain them. Thanks for understanding!

Issue #31, opened 2023-07-22
Add inputs, outputs controllers and position to hardware interfaces in application schema


eeberhard:
Similar to component descriptions, controller descriptions should allow the abstraction of input and output topic definition in separate fields `inputs` and `outputs`. The full hardware interface description should also include an optional position, same as components, so that the position of the node on the graph can be persisted.

Example:
```yaml
hardware_interfaces:
  robot:
    urdf: robot.urdf
    rate: 500
    position:
      x: 100
      y: 200
    controllers:
      my_controller:
        plugin: modulo_controllers/MyController
        parameters:
          foo: 1
        inputs:
          command: /my_command_topic
        outputs:
          state: /my_state_topic
```

Also, an optional `display_name` would be nice for the hardware interface.

PR #32, opened 2023-07-23
Extend application schema for controllers

eeberhard:
## Description

- #31 

Herein I add the extra fields to the application schema for hardware interfaces and controllers, namely a `position` and `display_name` for the hardware interface, and `inputs` and `outputs` for the controllers.

To reduce code duplication, I defined subschemas for inputs, outputs and parameters in a `common` directory, and moved the existing `position` schema there too. Then I used the same inputs, outputs and parameters in the controller schema as for the components. 

I also made the top-level `components` field in the application schema not required anymore. It was a bit weird being the only required element, since technically you could have an app that is just pure hardware interfaces. I think it comes from the days when we only had components and no hardware.

## Review guidelines

## Related issues

- https://github.com/aica-technology/dynamic-state-engine/issues/120

### Blocked by
- https://github.com/aica-technology/frontend/pull/268


domire8 (approved):
Looks good! Is it correct that `display_name` doesn't have a paragraph in the README and is that desired?

eeberhard:
> Looks good! Is it correct that `display_name` doesn't have a paragraph in the README and is that desired?

Hm nice catch, I missed that, not super serious but I'll add it anyway once resolve the upstream blocking PRs

PR #34, opened 2023-07-31
feat(docs): add docusaurus setup for docs

LouisBrunner:
## Description

See here: https://docs.aica.tech/

This replaces the Swagger documentation with a Docusaurus setup. I will readd the Swagger definition in a second PR after this by pulling from the backend-api repo (instead of the current pushing pattern), this will reduce automation for update but improve versioning.

Note that both the docs and the blog components of Docusaurus are disabled so none of the example code left is actually generating anything. I left those as examples for anyone to be able to pick it up after me. If we don't want the blog we can just delete all the associated code without issue. In a similar vein, I left the React code to show "features" even though it's a bit excessive.

Most of the boilerplate is left as-is, I only edited basic things (e.g. icons, colors, CI).

## Review guidelines


domire8 (approved):
cool!

eeberhard (approved):

PR #35, opened 2023-07-31
feat(docs): add API documentation back

LouisBrunner:
## Description

See it here: https://docs.aica.tech/api/

Readd the API removed in #34

The documentation is built in-step with the Backend API and then pinned in this repository. The `docs/api.version` file points to the version to be deployed (currently using a feature branch due to the lack of release with this feature).

## Review guidelines


LouisBrunner:
> I guess you need to change the branch but otherwise good

For the API version you need? We need a release then

domire8:
We could do that after https://github.com/aica-technology/backend-api/pull/85 no? @eeberhard

eeberhard:
> We could do that after [aica-technology/backend-api#85](https://github.com/aica-technology/backend-api/pull/85) no? @eeberhard

Yes thanks, that's partly why I was holding off on approval. I will get that PR resolved today so that we can release the backend API and then get this in.

docs/api.version
```diff
@@ -0,0 +1 @@
+feature-docs-46
```
domire8, line 1:
```suggestion
v1.0.0
```

PR #38, opened 2023-08-03
Deploy python API client to PyPi automatically

jjenscodee:
## Description

- #37 

Deploy python API client to PyPi automatically when on push to main, develop, or on a version tag, and only on changes to the `python` path.

An example of the workflow: https://github.com/aica-technology/api/actions/runs/5748364572
## Review guidelines


eeberhard (approved):
Good work, the result of the action and the output https://pypi.org/project/aica-api/0.0.5/ shows that it is working, and the workflow triggers are good as we discussed.

PR #43, opened 2023-08-25
feat: extend trigger event button's event type

jjenscodee:
## Description

- https://github.com/aica-technology/frontend/issues/345

This PR extends the event type for the trigger event button to support all event types.

## Review guidelines

### DIFF
--- a/schemas/applications/schema/buttons.schema.json
+++ b/schemas/applications/schema/buttons.schema.json
@@ -22,9 +22,6 @@
         "unload": {
           "$ref": "events/unload.schema.json"
         },
-        "transition": {
-          "$ref": "events/transition.schema.json"
-        },
         "service": {
           "$ref": "events/service.schema.json"
         },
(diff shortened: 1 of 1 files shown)


domire8 (approved):
Looks goot to me

schemas/applications/schema/buttons.schema.json
```diff
@@ -21,6 +21,30 @@
         },
         "unload": {
           "$ref": "events/unload.schema.json"
+        },
+        "transition": {
+          "$ref": "events/transition.schema.json"
+        },
```
eeberhard, line 27: this one makes no sense for the button and can be removed. 

Transition is designed as a shortcut between components. If component A triggers a transition to component B, then it will unload component A and load component B. But, if a button is connected to component B, it cannot "transition" since we cannot "unload" the button.
```suggestion
```

eeberhard (commented):
Good with me after a small change.

As a note, there is also a `set` event which is used to set component parameters. However, we don't need the trigger button to achieve this, since modifying the parameters in the component node block in the graph is the main way to set parameters. No extra button needed.

eeberhard (approved):

PR #44, opened 2023-08-26
feat(application schema): add support for setting controller parameters

domire8:
## Description

- https://github.com/aica-technology/dynamic-state-engine/pull/151

This PR adds the controller parameters to the `set` event in the schema. Because neither `controller/interface` nor `component` are required, I had to add basically three different objects that are allowed for the `set`. There might be a simpler version that I didn't see but this works and is tested (with vs code)

## Review guidelines

### DIFF
--- a/schemas/applications/schema/events/set.schema.json
+++ b/schemas/applications/schema/events/set.schema.json
@@ -39,29 +39,13 @@
       }
     },
     "controller": {
-      "title": "Conntroller Name", 
+      "title": "Controller Name", 
       "description": "The name of a controller",
       "type": "string",
       "pattern": "(^[a-zA-Z][a-zA-Z0-9_]*[a-zA-Z0-9]$)|(^[a-zA-Z]$)"
     },
     "set_parameter_object": {
       "oneOf": [
-        {
-          "type": "object",
-          "additionalProperties": false,
-          "properties": {
-            "parameter": {
-              "$ref": "#/$defs/parameter"
-            },
-            "value": {
-              "$ref": "#/$defs/value"
-            }
-          },
-          "required": [
-            "parameter",
-            "value"
-          ]
-        },
         {
           "type": "object",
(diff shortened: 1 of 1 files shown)


schemas/applications/schema/events/set.schema.json
```diff
@@ -15,40 +15,97 @@
...
+      "items": {
+        "type": [
+          "boolean",
+          "number",
+          "string"
+        ]
+      }
+    },
+    "controller": {
+      "title": "Conntroller Name",
```
eeberhard, line 42:
```suggestion
      "title": "Controller Name", 
```

schemas/applications/schema/events/set.schema.json
```diff
@@ -15,40 +15,97 @@
...
+            },
+            "value": {
+              "$ref": "#/$defs/value"
+            }
+          },
+          "required": [
+            "parameter",
+            "value"
+          ]
         },
```
eeberhard, line 64: Instead of having two definitions for the component parameter (one without a component, and one with), just one should suffice. The component field is not required, so just define it but don't include it in the "required" list
```suggestion
```

schemas/applications/schema/events/set.schema.json
```diff
@@ -15,40 +15,97 @@
...
+              "$ref": "#/$defs/value"
+            },
+            "component": {
+              "$ref": "identifiers/component.schema.json"
+            }
+          },
+          "required": [
+            "parameter",
+            "value",
+            "component"
```
eeberhard, line 82:
```suggestion
            "value"
```

schemas/applications/schema/events/set.schema.json
domire8: Yes, true. Thanks

jjenscodee (approved):

eeberhard (approved):

PR #48, opened 2023-10-02
feat(packaging): add package.sh to create a customer image using a license as authentication

LouisBrunner:
## Description

Add `package.sh` script to compose a customer (final) image using our base image and a various amount of other packages.

All images are hosted in our private Docker registry whose authentication is controlled by the provided license.

## Review guidelines

### DIFF
--- a/scripts/package.sh
+++ b/scripts/package.sh
@@ -1,0 +1,117 @@
+#!/bin/bash
+
+set -e
+
+BASE_TAG=""
+LICENSE_FILE="./License"
+PACKAGES=()
+FINAL_TAG=""
+REGISTRY="registry.licensing.aica.tech"
+LOCAL_IMAGES=no
+BASE_IMAGE="base"
+
+usage() {
+  if [ -n "$1" ]; then
+    echo "Error: $1"
+  fi
+  echo "Usage: $0 [options]"
+  echo "Options:"
+  echo "  --base <version>               base image version"
+  echo "  --packages <pkg1> <pkg2>       packages to install (e.g. 'collections/ur' or 'components/gear-assembly')"
+  echo "  --license <file>               license file to use to connect to the registry (default: '${LICENSE_FILE}')"
+  echo "  --tag <tag>                    docker tag to use for the final image"
+  echo "  -h, --help                     display this help"
+  if [ -n "${DEBUG_CLI}" ]; then
+    echo "Debugging options:"
+    echo "  --registry <registry>          registry to use (default: '${REGISTRY}')"
+    echo "  --local-images                 use local images instead of the registry"
+    echo "  --base-image <tag>             override the base image name and tag (requires --local-images)"
+    echo "  --                             stop parsing options, pass the rest to Docker build"
+  fi
+  exit 1
+}
+
+enforce_args() {
+  if [ "$#" -lt "$1" ]; then
+    NUM=$(( $1 - $# ))
+    usage "missing ${NUM} argument(s) for $2"
+    exit 1
+  fi
+}
+
+make_image_name() {
+  local image_with_tag="$1"
+  local image="$2"
+  if [ "${LOCAL_IMAGES}" = "yes" ]; then
+    echo "${image}"
+  else
+    echo "${REGISTRY}/${image_with_tag}"
+  fi
+}
+
+main() {
+  while [ "$#" -gt 0 ]; do
+    case "$1" in
+      --base) enforce_args 2 "$1"; BASE_TAG="$2"; shift 2;;
+      --license) enforce_args 2 "$1"; LICENSE_FILE="$2"; shift 2;;
+      --tag) enforce_args 2 "$1"; FINAL_TAG="$2"; shift 2;;
+      --registry) enforce_args 2 "$1"; REGISTRY="$2"; shift 2;;
+      --local-images) LOCAL_IMAGES=yes; shift;;
+      --base-image) enforce_args 2 "$1"; BASE_IMAGE="$2"; shift 2;;
+      -h|--help) usage; shift;;
+      --packages)
+        enforce_args 2 "$1"
+        shift
+        # until we run into something that looks like an option
+        while [ "$#" -gt 0 ]; do
+          case "$1" in
+            -*|--*) break;;
+          esac
+          PACKAGES+=("$1")
+          shift
+        done
+        ;;
+      --) shift; break;;
+      *) break;;
+    esac
+  done
+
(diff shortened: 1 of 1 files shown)


scripts/package.sh
```diff
@@ -0,0 +1,117 @@
...
+    echo "Error: $1"
+  fi
+  echo "Usage: $0 [options]"
+  echo "Options:"
+  echo "  --base <version>               base image version"
+  echo "  --packages <pkg1> <pkg2>       packages to install (e.g. 'collections/ur' or 'components/gear-assembly')"
+  echo "  --license <file>               license file to use to connect to the registry (default: '${LICENSE_FILE}')"
+  echo "  --tag <tag>                    docker tag to use for the final image"
+  echo "  -h, --help                     display this help"
+  if [ -n "${DEBUG_CLI}" ]; then
```
domire8, line 24: DEBUG_CLI is never set, whats the intended behavior here?

domire8 (commented):
Currently, our components and collections are build with the name `ghcr.io/aica-technology/collections/ur-collection:latest`. Using this package script to package with `--local-images` do I have to provide the full image name, or can I just do `./package.sh --local-images --packages collections/ur-collection:latest` or do I have to with full name?

LouisBrunner:
> Currently, our components and collections are build with the name `ghcr.io/aica-technology/collections/ur-collection:latest`. Using this package script to package with `--local-images` do I have to provide the full image name, or can I just do `./package.sh --local-images --packages collections/ur-collection:latest` or do I have to with full name?

You would need to add the full names as it would allow you to use one built locally as well (which I do a let personally).

scripts/package.sh
LouisBrunner: It's to be used by developers like us, by adding `DEBUG_CLI=anything` before calling the tool (you can export it too if you prefer).

scripts/package.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash
+
+set -e
+
+BASE_TAG=""
+LICENSE_FILE="./License"
+PACKAGES=()
+FINAL_TAG=""
```
eeberhard, line 8: Don't you think it would be more convenient to have some default tags? What the default final tag should / could be is up to debate, but leaving it empty also seems cumbersome. You've been using `customer:v1` in your examples, but a generic name for documentation, examples and the default here could be `aica-technology/app`, potentially with the base tag added.

Also nitpick, up until now the default license file has been lowercased.

```suggestion
BASE_TAG="latest"
LICENSE_FILE="./license"
PACKAGES=()
FINAL_TAG="aica-technology/app:${BASE_TAG}"
```

eeberhard (commented):
I don't want to suggest too many features to overcomplicate this script, but a main one I think would help is the option to mix local and registry images. For example, I might want a locally built base image with ECR extras, or conversely an ECR base with local extras, or finally a mix of extra packages that are either ECR or local.

To me it would make the most sense if `--packages` accumulated a list of packages on the registry, and then `--custom-packages` (a.k.a. `--local-images`)  accumulated a list of local images. `--packages` of course would take be the ECR package name, with the registry prefix automatically added, while `--custom-packages` takes the full image name which can be from anywhere.

I guess you would then have two loops, `for package in "${PACKAGES[@]}"` and `for custom_package in "${CUSTOM_PACKAGES[@]}"`, but then this would also remove the need for `make_image_name`.

And for the question of base image, you could have mutually exclusive `--base` (tag from ECR) or `--base-image` (fully namespaced image which could be from anywhere, including local build). 

Usage would look something like: 

```shell
# build with ECR base image
--base v1.1.0 --packages pkg1 pkg2 [--custom-packages pkg3 pkg4]

# build with local base image
--base-image aica-technology/base:debug --packages pkg1 pkg2 [--custom-packages pkg3 pkg4]
```

LouisBrunner:
> I don't want to suggest too many features to overcomplicate this script, but a main one I think would help is the option to mix local and registry images. For example, I might want a locally built base image with ECR extras, or conversely an ECR base with local extras, or finally a mix of extra packages that are either ECR or local.

That's basically what `--local-images` does, we could rename it as `--any-images` or something similar but I am not convinced that it's particularly more intuitive to have 2 sets of flags to do local vs registry. Also what if we want to use our private registry (GHCR)? I think having a way to turn off the registry prefixing and let you put any fully qualified image name is much more flexible.

scripts/package.sh
LouisBrunner: Actually we don't have a `latest` tag for `base` in the registry (or for any other images either by the way) as it makes quite hard to do licensing check with it so that's no possible and I will revert that part of the commit.

eeberhard:
I think it's great to automatically prefix the package names with the registry URI, and I understand the flexibility of disabling that automatic prefix. But I can see a common use case where I want the registry packages (for example, `collections/ur-collection` and `components/task-optimization`) but also some local `custom-component-package:v2` at the same time. It just seems verbose to have to supply the full registry name for "official" packages if I want to include any one local package

LouisBrunner:
> I think it's great to automatically prefix the package names with the registry URI, and I understand the flexibility of disabling that automatic prefix. But I can see a common use case where I want the registry packages (for example, `collections/ur-collection` and `components/task-optimization`) but also some local `custom-component-package:v2` at the same time. It just seems verbose to have to supply the full registry name for "official" packages if I want to include any one local package

But we also see that @domire8 is also interested in a `ghcr.io/aica-technology/` prefix, does that mean we need 3 pair of flags?

scripts/package.sh
eeberhard: Ah yes good point 👍

domire8:
> But we also see that @domire8 is also interested in a `ghcr.io/aica-technology/` prefix, does that mean we need 3 pair of flags?

I am indeed interested in that, you got that correctly :smile:

LouisBrunner:
> The option of swapping the registry prefix with --registry ghcr.io/aica-technology comes to mind, it would just need some logic around the login step.

Even this step is not going to be trivial IMO. Also what if you want both ECR and GHCR for whatever reason (e.g. some CI failure and the image doesn't exist in one)?

In any case, I was focusing on making a script that is easy to use for customers while having all the options for developers to bypass anything if required, which I think it does.

We can add all these options but I am not clear what that currently look like in terms of options and how complex the final script will be. I would say the next step would be for you to provide what the usage of such a script look like (the list of options and their arguments) and then I can see how much shellscript is required.

LouisBrunner:
Closed in favor of `app-builder`

PR #49, opened 2023-10-04
ci: update workflow before deleting develop branch

jjenscodee:
## Description

This PR updates the ci workflow before fast forward to `main` and delete the `develop` branch.

## Review guidelines

Estimated Time of Review: 1 minutes

## Checklist before merging:

### DIFF
--- a/.github/workflows/generate-ts-declaration.yml
+++ b/.github/workflows/generate-ts-declaration.yml
@@ -3,7 +3,6 @@ run-name: generate-typescript-declaration
 on:
   push:
     branches:
-      - develop
       - main
     paths:
       - '**.schema.json'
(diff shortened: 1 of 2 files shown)


.github/workflows/generate-ts-declaration.yml
```diff
@@ -41,7 +40,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: aica-technology/frontend
-          ref: develop
+          ref: main
```
domire8, line 43: Interesting, did that workflow never run since we deleted develop on frontend or did we not realized that it fails?

domire8 (approved):
Nice, thanks!

.github/workflows/generate-ts-declaration.yml
jjenscodee: this workflow runs only when the schemas change. maybe we haven't modify the schemas since we delete the branch in frontend(?

domire8: ah i see yeah that's it then probably

PR #52, opened 2023-10-09
fix(docs): base url for docusaurus

eeberhard:
## Description

I noticed that the GitHub docs pages (https://aica-technology.github.io/api) is not rendering correctly after the merge to main, and https://docs.aica.tech does not resolve to a valid GitHub page at all.

For the first issue, the suggested resolution is to be to fix up the base url. Not sure if that will resolve the second issue @LouisBrunner @jjenscodee ?

## Review guidelines

## Checklist before merging:

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -12,7 +12,7 @@ const config = {
   favicon: "img/favicon.ico",
 
   url,
-  baseUrl: "/",
+  baseUrl: "/api/",
 
   // GitHub pages deployment config.
   organizationName: "aica-technology",


LouisBrunner:
No, you can't do that as it means https://docs.aica.tech/ won't work (because it's base url is `/`).

I will look into why this link isn't working but I have seen that it gets unset sometimes in the settings of the repository (no idea why, looks like a GitHub bug to me).

LouisBrunner:
I fixed it (again), I genuinely don't know why it keeps being unset and I would seriously consider opening a GitHub support ticket to get this cleared out.

If it needs fixing again, here is the process:

* Go to this repository's settings
* Go to Pages
Screenshot 2023-10-09 at 11 42 57

* The settings for custom domain will be cleared for some unknown reason
Screenshot 2023-10-09 at 11 42 28

* Re add the domain and click save
Screenshot 2023-10-09 at 11 42 53

eeberhard:
Thanks for the info!

Issue #55, opened 2023-10-12
Endpoint for check()


domire8:
In the python API client, there is a `check()` method that gets `/v1/` but I don't think that this endpoint even exists (anymore). Should we change that to `/v1/application/state` or just `/version` ?

@eeberhard @jjenscodee

jjenscodee:
what happens if client version != api version?

PR #60, opened 2023-10-18
fix: update check function

jjenscodee:
## Description

- #55 

This PR updates the `check()` function to check if the API version is at least v2.

## Review guidelines

## Checklist before merging:

### DIFF
--- a/python/src/aica_api/client.py
+++ b/python/src/aica_api/client.py
@@ -2,7 +2,7 @@ from typing import Union, List
 
 import requests
 
-from aica_api.ws_client import WebsocketSyncClient
+from ws_client import WebsocketSyncClient
 
 
 class AICA:
@@ -53,7 +53,10 @@ class AICA:
         """
         # TODO: come up with a compatibility table in the future
         api_version = requests.get(f'{self._address}/version').json()
-        return api_version.startswith('2')
+        new_version = api_version.startswith('2')
+        if not new_version:
+            print('You are using an old version ' + api_version)
+        return new_version
 
     def component_descriptions(self) -> requests.Response:
         """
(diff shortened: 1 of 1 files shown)


python/src/aica_api/client.py
```diff
@@ -47,11 +47,16 @@ def _ws_endpoint(self, endpoint=''):
...
 
-    def check(self) -> requests.Response:
+    def check(self) -> bool:
         """
-        Make a GET request to the default endpoint to verify connectivity.
+        Check if the API version is v2 (any v2.x.x tag)
         """
-        return requests.get(self._endpoint())
+        # TODO: come up with a compatibility table in the future
+        api_version = requests.get(f'{self._address}/version').json()
```
eeberhard, line 55: suggestion: I would also add a try / catch around this line, that way we can also check for connection errors (i.e. API server is not running or is on a different port / address)

```python
try:
    api_version = requests.get(f'{self._address}/version').json()
except requests.exceptions.RequestException as e:
    print(f'Error connecting to the API! {e}')
    return False
```

python/src/aica_api/client.py
```diff
@@ -47,11 +47,16 @@ def _ws_endpoint(self, endpoint=''):
...
+        new_version = api_version.startswith('2')
+        if not new_version:
+            print('You are using an old version ' + api_version)
```
eeberhard, line 58: wording suggestion
```suggestion
            print(f'The detected API version v{api_version} is older than the minimum API version v2.0.0 supported by this client')
```

domire8 (approved):

python/src/aica_api/client.py
```diff
@@ -2,7 +2,7 @@
 
 import requests
 
-from aica_api.ws_client import WebsocketSyncClient
```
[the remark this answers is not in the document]
eeberhard: Oof it seemed suspicious at the time, I should have checked that, sorry!

Issue #59, opened 2023-10-18
Update application schema with unified syntax


domire8:
Given all the changes that are tracked by the parent issue, the application syntax needs to be updated too.

domire8:
Here, I'm not sure how you want to do it @eeberhard. On one hand we are now backwards compatible with the syntax that is currently described here. On the other hand we already support the new event syntax. Do you want to
- Tag the current version of this syntax as version 1, or
- Already add the syntax here and have a version 1 where both the old and new syntax are described?

I find option 2 a bit tricky because we'll put in work that is not strictly necessary if we know that version 2 is just around the corner.

eeberhard:
No, I would like to modify the current schema to only describe the new unified syntax, and then tag that as schema version 1

Backwards compatibility with "pre-release" syntax is only to support our existing applications and demos, but should not be officially supported with a version or documentation moving forward.

Note: I would like to use [SchemaVer](https://docs.snowplow.io/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver/) to track json schemas, both because it is semantically different from semver and so as not to get mixed up between API version or python client version. A corresponding GitHub tag might then be `applications-1-0-0`. The specifics around that will be decided and fixed in the final steps of this refactoring work

PR #61, opened 2023-10-18
refactor: update application schema with unified syntax

domire8:
## Description

- #59

This PR solves the issue by revising the schema with the most recent changes in the DSE.

## Review guidelines

## Checklist before merging:


schemas/applications/schema/events/set.schema.json
```diff
@@ -75,17 +69,17 @@
             "value": {
               "$ref": "#/$defs/value"
             },
-            "interface": {
+            "hardware": {
               "$ref": "identifiers/hardware.schema.json"
             },
             "controller": {
-              "$ref": "#/$defs/controller"
+              "$ref": "identifiers/controller.schema.json#/properties/controller"
```
eeberhard, line 76: good use of json pointer reference

schemas/applications/schema/events/load.schema.json
```diff
@@ -1,18 +1,51 @@
...
+              "$ref": "identifiers/hardware.schema.json"
+            }
+          },
+          "required": [
+            "hardware"
+          ]
+        },
+        {
+          "$ref": "identifiers/controller.schema.json"
+        }
```
eeberhard, line 47: I was trying to think of a more compact way to define these three cases (since it's also repeated in `unload`, but because the component and hardware identifiers are just strings, you do indeed need to define an object around each of them, so I think it's good 👍

eeberhard (commented):
No blocking comments except to think about changing the top-level `hardware_interfaces` field to `hardware`. I would be for making that change (in the DSE yaml parser, it would be basically a one-line change with backwards compatibility as before)

(also I did this review a few days ago but turns out I never submitted it 😑

schemas/applications/schema/application.schema.json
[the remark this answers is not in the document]
domire8: Yeah I would like that too :+1: I thought about it but hesitated to bring it up

eeberhard (approved):

PR #63, opened 2023-10-20
docs: update docs after syntax changes

domire8:
## Description

While I was implementing #61, I always tried to address the same problems in the docs too. Here are all those changes that I came up with so far.

## Review guidelines

## Checklist before merging:


docs/docs/getting-started/04-timer-example.md
eeberhard, line 1: Don't forget to change the `on_start` in the example from
```yaml
on_start:
  load: timer
```
to
```yaml
on_start:
  load:
    component: timer
```

eeberhard (approved):

docs/docs/getting-started/04-timer-example.md
domire8: Good catch thanks

PR #70, opened 2023-10-31
fix: revert import

jjenscodee:
## Description

(https://github.com/aica-technology/api/pull/60#discussion_r1377726768)

This PR solves the issue by updating the import. It also updates the version for api client.

## Review guidelines

## Checklist before merging:


python/CHANGELOG.md
```diff
@@ -2,8 +2,12 @@
 
 Release Versions:
 
+- [1.0.1](#70)
```
domire8, line 5:
```suggestion
- [1.0.1](#101)
```

JasperTan97 (approved):
lgtm

python/CHANGELOG.md
```diff
@@ -2,8 +2,12 @@
...
 - [1.0.0](#100)
 
+## 1.0.1
+Version 1.0.1 fixes the import issue.
```
domire8, line 9:
```suggestion

Version 1.0.1 fixes a relative import issue.
```

domire8 (approved):
After that its okay for me, but @eeberhard knows more about relative imports in python

eeberhard (approved):
Since the python package is called `aica_api`, all modules need to be imported from the package namespace. The other change may have worked for local testing because the module was resolved on the python path, but that would not be the case after the package is installed to site packages.

Issue #62, opened 2023-10-19
Track JSON schema versions


eeberhard:
The JSON schemas should be versioned according to [SchemaVer](https://docs.snowplow.io/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver/), starting at 1-0-0.

But, how to associate specific JSON schemas with a version?

## Option 1

Require the `version` property at the top level of the schema object, and validate it against a const string value.

```json
   "required"   : ["version"],
   "properties" : {
     "version" : { "type":"string", "const":"1-0-0" },
     ...
   }
```

### Pros
- schema version is defined in source, so it doesn't depend on file name / path / URI
- schema version must also be defined in the JSON data itself to be correctly validated (encouraging us to track syntax versions in applications and descriptions, and making it easier to spot outdated apps)

### Cons
- schema version must also be defined in the JSON data itself to be correctly validated (which would require updating existing json data)
- does not support additions or revisions without additional work, e.g. `"enum": ["1-0-0", "1-0-1", "1-1-0"]`

## Option 2

Make a git tag with `<schema_name>_<schema_version>`, e.g. `applications_1-0-0`.

### Pros
- possible to find specific schema versions using tags or URLs 

### Cons
- many tags at the repo level (one for each schema and version), could be confusing if you don't know about the versioning system

## Option 3

Use the top-level `"description"` property of the schema to track the schema version

### Pros
- schema version is defined in source
- super simple, human readable

### Cons
- harder to use for automatic version checking without custom parsing of the schema document
- prevents normal use of description field unless there is some consistent convention (i.e. description always starts with version, semicolon, then text description)

## Option 4

Use the `"$id"` property of the schema to define the base URI with a version identifier

### Pros
- schema version is defined in source
- makes cross-referencing of schemas and subschemas more explicitly version-locked

### Cons
- potential mismatch between specified base URI and actual retrieval URI

eeberhard:
Also to think about: 
 
- Hosting versioned schemas in more easily retrievable place (i.e. rather than permalinking to GitHub source, host on `docs.aica.tech/schemas/...`). This could be part of the workflow in API repository.
- Using bundling tools to compile / flatten schemas and subschemas into one file. This would make hosting and retrieval of specific schemas easier, and may also simplify versioning as a result

eeberhard:
@LouisBrunner @domire8 would be happy for your thoughts around this issue

domire8:
[...]

PR #73, opened 2023-11-13
ci: track json schema versions

jjenscodee:
## Description

- #62 

This PR adds the "$id" property of the schema to define the base URI with a version identifier and introduce a new workflow to:
1. looks for changes to the version in the $id of the top-level schema
2. bundles the schemas into a single flat
3. push the file to the `docs/static/schemas` under a path that matches the retrieval URI of the top-level $id

## Review guidelines

Estimated Time of Review: 8 minutes

## Checklist before merging:


.github/actions/bundle-schema/action.yml
```diff
@@ -0,0 +1,53 @@
...
+      if: ${{ steps.versions.outputs.new_version != steps.versions.outputs.prev_version }}
+      shell: bash
+      run: |
+        python3 utils/bundle.py ${{ inputs.schema-path }} ${{ inputs.schema }}
+        mkdir -p docs/static/schemas/${{ steps.versions.outputs.version_only }}
+        mv ${{ inputs.schema }} docs/static/schemas${{ steps.versions.outputs.new_version }}
+        git config --global user.name github-actions[bot]
+        git config --global user.email github-actions[bot]@users.noreply.github.com
+        git add .
+        git commit -m "generate new schema"
```
eeberhard, line 52:
```suggestion
        git commit -m "docs: ${{ inputs.schema }} ${{ steps.versions.outputs.version_only }}"
```

schemas/component-descriptions/schema/component.schema.json
eeberhard, line 1: could you also delete the old id tags from the subschemas in this directory? the `predicate`, `service` and `signal_collection` sub-schemas don't need their own id afterwards.

.github/workflows/bundle-schema.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Bundle & Push JSON schema for application schemas
+        uses: ./.github/actions/bundle-schema
+        with:
+          schema-path: ./schemas/applications/schema/application.schema.json
+          schema: application.schema.json
+      - name: Bundle & Push JSON schema for component-descriptions schemas
+        uses: ./.github/actions/bundle-schema
+        with:
+          schema-path: ./schemas/component-descriptions/schema/component.schema.json
+          schema: component.schema.json
```
eeberhard, line 26: Could you also update the `$id` of the controller description schema and include it here?

```suggestion
          schema: component.schema.json
      - name: Bundle & Push JSON schema for controller-descriptions schemas
        uses: ./.github/actions/bundle-schema
        with:
          schema-path: ./schemas/controller-descriptions/schema/controller.schema.json
          schema: controller.schema.json
```

jjenscodee:
> Just to check, the bundle-schema action will bundle and push commits directly to main right? 

Yes, that's what I intend to do

.github/actions/bundle-schema/action.yml
```diff
@@ -0,0 +1,53 @@
...
+      shell: bash
+      run: |
+        python3 utils/bundle.py ${{ inputs.schema-path }} ${{ inputs.schema }}
+        mkdir -p docs/static/schemas/${{ steps.versions.outputs.version_only }}
+        mv ${{ inputs.schema }} docs/static/schemas${{ steps.versions.outputs.new_version }}
+        git config --global user.name github-actions[bot]
+        git config --global user.email github-actions[bot]@users.noreply.github.com
+        git add .
+        git commit -m "docs: ${{ inputs.schema }} ${{ steps.versions.outputs.version_only }}"
+        git push
```
LouisBrunner, line 53: It pushes on the same branch?

jjenscodee: yes it pushes on the same branch. There's a condition for the workflow:
`
on:
  push:
    branches:
      - main
`
so I think the workflow happens on the main branch and this push will push the changes to main, maybe I'm wrong? @LouisBrunner

LouisBrunner: Yes, that's correct! I missed the `on` condition, thanks.

I don't like when the CI pushes on branches used by humans as it is a pain to use (have to pull everytime you push, etc) but it's a good v1 and we are happy with the downsides for now I think.

LouisBrunner (approved):

.github/actions/bundle-schema/action.yml
eeberhard: yes I think the benefits of this approach outweigh the downsides and it's a good v1

PR #77, opened 2023-11-17
fix: bundling issues with application schema

jjenscodee:
## Description

- #78 

I update the `bundleHelper.js` to avoid some symbols being replaced. Also remove the old "id" I found in other files.

## Review guidelines

## Checklist before merging:

### DIFF
--- a/utils/bundleHelper.js
+++ b/utils/bundleHelper.js
@@ -4,10 +4,8 @@ import * as fs from "fs";
 let filePath = process.argv[2];
 let outputFileName = process.argv[3];
 $RefParser.bundle(filePath).then(async (schema) => {
-    let jsonSchema = JSON.stringify(schema, null, 2);
+    let jsonSchema = decodeURI(JSON.stringify(schema, null, 2));
     jsonSchema = jsonSchema.replaceAll('%24', '$');
-    jsonSchema = jsonSchema.replaceAll('(%5E%5Ba-zA-Z%5D%5Ba-zA-Z0-9_%5D*%5Ba-zA-Z0-9%5D$)%7C(%5E%5Ba-zA-Z%5D$)', '(^[a-zA-Z][a-zA-Z0-9_]*[a-zA-Z0-9]$)|(^[a-zA-Z]$)');
-
     fs.writeFile(outputFileName, jsonSchema, 'utf8', (err) => {
         if (err) throw err;
         console.log('The file has been saved!');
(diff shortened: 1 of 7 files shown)


jjenscodee:
This PR is mainly for the application schema. Still trying to figure out with the component & controller.

utils/bundleHelper.js
```diff
@@ -3,14 +3,15 @@ import * as fs from "fs";
 
 let filePath = process.argv[2];
 let outputFileName = process.argv[3];
-$RefParser.bundle(filePath).then((schema) => {
+$RefParser.bundle(filePath).then(async (schema) => {
     let jsonSchema = JSON.stringify(schema, null, 2);
     jsonSchema = jsonSchema.replaceAll('%24', '$');
+    jsonSchema = jsonSchema.replaceAll('(%5E%5Ba-zA-Z%5D%5Ba-zA-Z0-9_%5D*%5Ba-zA-Z0-9%5D$)%7C(%5E%5Ba-zA-Z%5D$)', '(^[a-zA-Z][a-zA-Z0-9_]*[a-zA-Z0-9]$)|(^[a-zA-Z]$)');
```
eeberhard, line 9: The issue lies with the percent-encoding (URI encoding) of these symbols, which as you noticed already happens with the dollar sign. Rather than replacing a specific text block (and then potentially having the same problem again if we change it or add new symbols), I think it would be better to decode all percent-encoded symbols. You could do it with separate calls to replace occurrences of `%5B` with `[`, `%5D` with `]`, `%5E` with `^` and `%7C` with `*`. Or, I think you can just use the built-in function `decodeURI()` on the whole jsonSchema string...

```suggestion
    let jsonSchema = decodeURI(JSON.stringify(schema, null, 2));
```

jjenscodee: I just tried with this decodeURI and seems like it won't work if we have the '$' case. Still gets the '%24'.

eeberhard: I guess the dollar sign is allowed in URI encodings (like hashtag, question mark, etc). Still it's nice to have the general decoding for the majority of the characters!

eeberhard (approved):

domire8:
I see you edited the files under `static` folder. Is that the hosted file, e.g. will then the one that we can access through the docs page be updated?

utils/bundleHelper.js
jjenscodee: yeah I think so but it didn't work with vscode when I try the file with '%24' so that's why I still keep the replace there.

jjenscodee:
> will then the one that we can access through the docs page be updated?

Yes, but only when on the main branch.

domire8 (approved):

PR #81, opened 2023-12-05
feat(client): support lifecycle transition

eeberhard:
## Description

A small feature that I didn't realize was missing until requested by @domire8, this adds the endpoint to set lifecycle transitions.

It also tags a new minor release v1.1.0 which includes the fix from #80.

## Review guidelines

## Checklist before merging:


domire8 (approved):
Thanks, this is tested!

Issue #82, opened 2023-12-08
Dereference schemas


jjenscodee:
Putting a "$schema" tag at the beginning of the file doesn't work in vscode because it cannot resolve nested references in the schema.

Issue #79, opened 2023-11-28
Desired usage of schemas


domire8:
@LouisBrunner @eeberhard @jjenscodee 

PR #78 fixed some issues with the schemas. However, we are still not sure where we are going with the schemas and what the desired usage is at the end. I will try to summarize what we have right now:

### What's possible

We can tell vscode to look up the schema from our docs page in a `settings.json` and we get autocomplete and validation.

### What's not possible

Putting a `"$schema"` tag at the beginning of the file doesn't work in vscode because it cannot resolve nested references in the schema.

### What's possible

We can use the YAML extension in vscode and configure the `settings.json` to look up the yaml syntax from the docs page.

### What's not possible

What do we want?

---

As far as I understand, it would be desirable to have the schema link be put directly inside the files that we are trying to validate because only like that can we 'control' what version is should be validated against. 
- For the component/controller descriptions, I think the problem with the nested reference is a problem that one can fix
- For the application, I'm really not sure how we should achieve that. Do you guys have an idea of what the goal would be for the YAML files?

jjenscodee:
From @eeberhard:
> At this point, the backend API now already does the job of “expanding” component descriptions by looking up the inheritance chain and building a new flat description for each component. We never actually use the nested reference for inheritance. Therefore, my suggestion would be to deprecate the nested inheritance property and just use this:
> ```
> "inherits": {
>   "$ref": "#/$defs/registration"
> }
> ```

eeberhard:
TODO: make a new epic containing these points:

https://github.com/aica-technology/api/issues/62#issuecomment-1775575486

domire8:
See #87

PR #83, opened 2023-12-08
fix: dereference schema

jjenscodee:
## Description

- #82 

This PR solves the issue by updating the ref path for the special case "parameter_state_type"

Related issue: https://github.com/aica-technology/api/issues/79

## Review guidelines

## Checklist before merging:


eeberhard:
Ah well this is not really so nice, mainly because `"$ref": "events.schema.json"` is used in about 10 places and itself references 7 subschemas, which amounts to around 10,000 lines of repeated schema.

I'm pretty sure the original issue (#79) was just for the nested reference in the component / controller description, because I can use the application schema as-is without. Really the ability for JSON to self-reference is very powerful and shouldn't be avoided by brute-force. The problem with dereferencing everything, apart from the file size being much bigger than it needs to be, it that it makes make the typescript generation basically unusable. Try it yourself:
```
curl https://raw.githubusercontent.com/aica-technology/api/fix/dereference-schema/docs/static/schemas/1-0-0/application.schema.json | json2ts -o test.d.ts
```

utils/bundleHelper.js
```diff
@@ -3,13 +3,25 @@ import * as fs from "fs";
...
-    });
+
+    if(outputFileName === "application.schema.json") {
+        fs.writeFile(outputFileName, JSON.stringify(jsonSchema, null, 2), 'utf8', (err) => {
+              if (err) throw err;
+              console.log('The file has been saved!');
+        });
+    } else {
+        let jsonSchemaObject = JSON.parse(jsonSchema);
+        jsonSchemaObject['properties']['parameters']['items']['properties']['parameter_state_type']['$ref'] = '#/properties/inputs/items/properties/signal_types/items';
```
eeberhard, line 19: nitpick: I would just suggest to add a comment here to explain why this is needed, for future reference

docs/static/schemas/1-0-0/application.schema.json
eeberhard, line 1: clearly something went wrong with the serialization / deserialization because you have all of the escaped `\n`and `\"` instead of expected format

eeberhard (commented):
You removed the inheritance ref from `schemas/component-descriptions/schema/component.schema.json`, could you please do the same for `schemas/controller-descriptions/schema/controller.schema.json`?

eeberhard (approved):
Nice work!

If I add `"$schema": "https://raw.githubusercontent.com/aica-technology/api/68d9ca5efc51ceb503ba4e8342491f365d879e9a/docs/static/schemas/1-0-0/component.schema.json"` to the top of a component description JSON file in VSCode, it seems to parse the schema correctly 🥳

domire8:
GitHub has problems, I can't click accept but well done! nice

Issue #84, opened 2023-12-08
Application schema for buttons is not correctly nested for multiple buttons


eeberhard:
As noted [here](https://github.com/aica-technology/frontend/blob/1d44f07150e2c172ebdd8e26f2e952975ce69076/src/utils/instances/application-instance-manager.tsx#L32-L36) after Front-End PR [#501](https://github.com/aica-technology/frontend/pull/501), the button schema is not correctly nested.

It currently defines the individual button properties (`on_click`, `position`, `display_name`) under the top-level Buttons object: 

```yaml
buttons:
  on_click:
    load: ...
```

But, the intended pattern is to have named buttons as part of the Buttons object:

```yaml
buttons:
  test_button:
    on_click:
      load: ...
  another_button:
    on_click:
      unload: ...
```

This is also how the backend and frontend parsers currently treat the button syntax.

PR #85, opened 2023-12-08
fix(schema): correctly nest buttons in application schema

eeberhard:
## Description

- #84 

This PR solves the issue by adding the pattern property for individual buttons, exactly the same as for components and hardware

## Review guidelines

## Checklist before merging:


eeberhard:
@jjenscodee after #83, would you mind rebasing this and re-running the bundle helper? As per my comment [here](https://github.com/aica-technology/api/issues/79#issuecomment-1847007800), I think we should also directly update the 1-0-0 static schema with this change rather than tagging a new version, since it's a "bug fix" more than a schema addition / revision.

jjenscodee (approved):

Issue #89, opened 2023-12-14
Set application from filepath


domire8:
Original discussion [here](https://github.com/JasperTan97/touch_localisation/pull/6#discussion_r1425606623)

As I've mentioned, I'm not sure there is a sensible solution for a helper that can easily detect if a string that ends with `.yaml` should be automatically opened as a file or not.

@eeberhard

eeberhard:
The original discussion is private. I will summarize it here.

## Feature suggestion

A common use of the python API client is to set a YAML application from a file. That currently looks like this:

```python
import yaml
from aica_api.client import AICA


aica = AICA()
my_app_filepath = '/path/to/my_app.yaml'
with open(my_app_filepath, 'r') as file:
    my_app_content = yaml.safe_load(file)
aica.set_application(my_app_content)
```

Potentially, it could be nice to do some logic inside of the `aica.set_application()` function to check if the argument is a filepath, and if the file exists it could read it and send then contact automatically. Then the usage would just be:

```python
from aica_api.client import AICA


aica = AICA()
my_app_filepath = '/path/to/my_app.yaml'
aica.set_application(my_app_filepath)
```

## Discussion

I first suggested looking for a `.yaml` extension.

Dominic's input:

> Detecting a .yaml extension is not good because we might actually want to send a filename to the computer running the app since the app might be on that computer and we happen to know the filepath on that computer (I do that regularly).

I forgot that the actual API endpoint also accepts a path to a file in the running application container. Dominic goes on to suggest:

> What we can do is see if the string ends with .yaml and is to be found on the machine making the api call, then we load it and otherwise we don't. But that might lead to confusing behavior

I think checking if a file exists, then loading and parsing it, and otherwise reverting to sending the raw input argument to the API, allows flexible and compact usage for a very common use-case. The only downside is potentially confusing behavior, but I think this can be avoided just by checking the conditions and printing some logs.

I still think it makes sense as an improvement.

PR #90, opened 2023-12-14
feat: parse yaml file if it exists

domire8:
## Description

- #89 

This issue suggests that we could log something, that would be 'new' for this client though. Should we just use `print()`?

## Review guidelines

## Checklist before merging:

### DIFF
--- a/python/src/aica_api/client.py
+++ b/python/src/aica_api/client.py
@@ -138,11 +138,9 @@ class AICA:
         """
         Set an application to be the current application.
 
-        :param payload: The filepath of an application on the AICA computer, or the application content as a
-        YAML-formatted string
+        :param payload: The filepath of an application or the application content as a YAML-formatted string
         """
         if payload.endswith(".yaml") and os.path.isfile(payload):
-            # FIXME print line?
             with open(payload, "r") as file:
                 payload = yaml.safe_load(file)
         data = {
--- a/python/src/aica_api/client.py
+++ b/python/src/aica_api/client.py
@@ -1,6 +1,8 @@
 from typing import Union, List
 
+import os
 import requests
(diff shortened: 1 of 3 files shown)


python/src/aica_api/client.py
```diff
@@ -139,6 +141,10 @@ def set_application(self, payload: str) -> requests.Response:
         :param payload: The filepath of an application on the AICA computer, or the application content as a
         YAML-formatted string
```
eeberhard, line 142: nitpick: this can now be simplified, since it covers all cases
```suggestion
        :param payload: The filepath of an application or the application content as a YAML-formatted string
```

Conversely you could be more verbose, as you prefer: "The filepath of an application on the host computer or the AICA application container, or the application content ..."

python/src/aica_api/client.py
```diff
@@ -139,6 +141,10 @@ def set_application(self, payload: str) -> requests.Response:
...
         """
+        if payload.endswith(".yaml") and os.path.isfile(payload):
+            # FIXME print line?
```
eeberhard, line 145: Fair point, adding a log output is a separate point and printing should be avoided. Let's leave it out for now.

eeberhard (approved):
I don't think there will be that much confusion on the usage. It's basically impossible that a valid YAML file also happens to match a real filepath. If do you supply a real filepath that is _not_ a YAML file, it will throw a `SyntaxError` which is the best way to handle it I think. The only subtle point is that it will prioritize local filepaths over container filepaths in case there is the exact same directory and file name pattern, which I also think is a very rare occurrence.

I would also directly tag this as a `1.1.0`, provided there are no pending features that we would want to bundle in that release.

Issue #94, opened 2024-01-09
Update python API client to use Socket.IO


eeberhard:
Use `python-socketio` client instead of websockets for the upcoming run-time data protocol refactor

PR #95, opened 2024-01-09
refactor!: use Socket.IO client instead of websockets for run-time data

eeberhard:
## Description

- https://github.com/aica-technology/api/issues/94

This PR solves the issue by replacing the websocket client with Socket.IO client. Breaking change, should be released as version 2.0.0 once the backend API is released as v3.0.0.

## Review guidelines

Estimated Time of Review: x minutes

## Checklist before merging:


LouisBrunner (approved):

JasperTan97 (approved):

python/src/aica_api/sio_client.py
```diff
@@ -0,0 +1,65 @@
...
+                    if elapsed >= timeout:
+                        raise TimeoutError
+                    received = sio.receive(timeout=timeout - elapsed)
+            except TimeoutError:
+                break
+            else:
+                if event == '*' or event == received[0]:
+                    try:
+                        if callback(data):
+                            return data
```
eeberhard, line 62:
```suggestion
                        if callback(received[1]):
                            return received[1]
```

jjenscodee (approved):

python/pyproject.toml
```diff
@@ -13,7 +13,7 @@ readme = "README.md"
 requires-python = ">=3.7"
 dependencies = [
     "requests >= 2.28.1",
-    "websocket-client >= 1.4.2"
+    "python-socketio >= 5.11.0"
```
eeberhard, line 16: From @JasperTan97's testing:
```suggestion
    "python-socketio[client] >= 5.11.0"
```

JasperTan97: Don't forget to commit the suggestion!

eeberhard: yes indeed. is it fine with you if I don't merge this PR until we actually release v3?

domire8:
Can this move forward as well now?

python/src/aica_api/client.py
```diff
@@ -40,30 +41,37 @@ def _endpoint(self, endpoint=''):
...
-        if not new_version:
-            print(f'The detected API version v{api_version} is older than the minimum API version v2.0.0 supported by '
-                  f'this client')
-        return new_version
+
+        if api_version.startswith('3'):
+            return True
+        elif api_version.startswith('2'):
+            print(f'The detected API version v{api_version} is older than the minimum API version v3.0.0 supported by '
+                  f'this client (v{self.client_version()}). Please install Python API client version v1.2.0 or later'
```
domire8, line 69: Is `later` the correct word here? I would understand it as > 1.2.0

domire8 (approved):
For me this still looks good, I will test again in the next couple of days but it can be merged before since the release will happen later

python/src/aica_api/client.py
eeberhard: Good point, since we know we aren't release anything between 1.2 and 2.0 we can make this more specific

eeberhard:
```suggestion
                  f'this client (v{self.client_version()}). Please install Python API client version v1.2.0 '
```

JasperTan97:
I've been using this since Jan 9 or 10 and had no issues with the API

PR #96, opened 2024-01-11
docs: correct typehints for setting parameters

domire8:
## Description

- https://github.com/aica-technology/backend-api/pull/130

This PR corrects the typehints for setting parameters

## Review guidelines

## Checklist before merging:


eeberhard (approved):
Thanks for catching that! It also occurred to me yesterday that the type hints were misleading here but I forgot to write it down and would have missed it for a while probably

PR #97, opened 2024-01-12
fix: required fields for hardware schema

domire8:
## Description

This PR solves a small issue in the hardware schema where the `required` field was in an incorrect place and didn't contain `urdf`.

I would like to release 1-0-1 version of the application schema now, so that raises several questions here @jjenscodee :
1. What do I need to do in addition to my changes such that the schema is generated?
2. Will a workflow create a new directory here https://github.com/aica-technology/api/tree/main/docs/static/schemas?
3. Will we have to manually edit the 1-0-1 files like we had to do for the 1-0-0 or is this now all inside a workflow?

Thanks for helping me out here.

## Review guidelines

## Checklist before merging:

### DIFF
--- a/docs/docs/reference/02-yaml-syntax.md
+++ b/docs/docs/reference/02-yaml-syntax.md
@@ -733,7 +733,7 @@ component:
 
 ## Validating a YAML application
 
-The [YAML application schema](https://docs.aica.tech/schemas/1-0-0/application.schema.json) defines the structural rules
+The [YAML application schema](https://docs.aica.tech/schemas/1-0-1/application.schema.json) defines the structural rules
 of an AICA application and effectively distinguishes between valid and invalid syntax.
 
 Many modern IDEs and code editors can be configured to support custom schemas and provide in-line validation and
@@ -747,5 +747,5 @@ Developers working with Visual Studio Code can validate YAML application files e
 2. Associate a schema with the YAML application by adding the following modeline to the file:
 
 ```yaml
-# yaml-language-server: $schema=https://docs.aica.tech/schemas/1-0-0/application.schema.json
+# yaml-language-server: $schema=https://docs.aica.tech/schemas/1-0-1/application.schema.json
 ```
--- a/schemas/applications/schema/application.schema.json
+++ b/schemas/applications/schema/application.schema.json
@@ -1,5 +1,5 @@
 {
-  "$id": "/1-0-0/application.schema.json",
+  "$id": "/1-0-1/application.schema.json",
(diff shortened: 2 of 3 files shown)


eeberhard:
For the release, technically all of your questions 1-3 should be covered just by changing the version in the `$id` of `application.schema.json`. The workflow will detect the version change, then bundle and push the new version to the docs.

This is the first time we are properly trying it after 1-0-0, which at the time required a few bug fixes and interventions, so let's see how it goes.

I doubled-check the SchemaVer convention (https://docs.snowplow.io/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver/) and actually the change here is not fully backwards compatible; we are making some properties required that previously were not required (at least, not explicitly required by the schema although the AICA framework would have still complained!). So this constitutes a `REVISION` and should be 1-1-0, not 1-0-1. I will have a quick look through the schema to see if there are any other minor changes we can / should bundle with this release.

Finally, could you add a CHANGELOG.md to the application schema directory to track these revision changes?

jjenscodee:
> What do I need to do in addition to my changes such that the schema is generated?

Change the id like you did -> this will trigger the workflow to bundle schema and push the result directly to main

> Will a workflow create a new directory here https://github.com/aica-technology/api/tree/main/docs/static/schemas?

Yes

> Will we have to manually edit the 1-0-1 files like we had to do for the 1-0-0 or is this now all inside a workflow?

The workflow is supposed to do everything if the version changes so no need to manually edit

jjenscodee (approved):

eeberhard:
The only thing that currently comes to mind is that we should probably specify `additionalProperties` to false in the top-level application object, maybe even everywhere else that is missing. I made some comments a while ago that I didn't see the point of adding that everywhere, because people might want to add some user data / metadata to the yaml apps. But as the [schemaver reference](https://docs.snowplow.io/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver/) points out, revising a schema to contain a new field _might_ conflict with an existing user app. The `additionalProperties: false` is then a way of "reserving" all other fields for future use. 

Apart from that I can think of a few additions (adding metadata for syntax version, expected base and package versions, and optional user-data field that can contain anything (which would still allow user extensions even with additionalProperties off). I don't know if any of those are worth blocking this 1-1-0 revision.

domire8:
If we don't have an urgent need for those minor adaptations, I would say it can wait. Otherwise we make a 1-2-0 soon, wouldnt be a big deal either.

eeberhard (approved):
Let's go for it, and keep an eye on the workflows to make sure they all pass as expected!

PR #106, opened 2024-01-15
ci: open a PR from a CI feature branch for bundle-schema action

jjenscodee:
## Description

- #100 

This PR updates the bundle-schema workflow to open a PR from a CI feature branch (ci/update-schemas)

testing result here: https://github.com/aica-technology/api/pull/109

## Review guidelines

Estimated Time of Review: 4 minutes

## Checklist before merging:

### DIFF
--- a/.github/actions/bundle-schema/action.yml
+++ b/.github/actions/bundle-schema/action.yml
@@ -49,6 +49,7 @@ runs:
         git config --global user.name github-actions[bot]
         git config --global user.email github-actions[bot]@users.noreply.github.com
         git fetch
+        git pull
         git checkout ci/update-schemas
         git add .
         git commit -m "docs: ${{ inputs.schema }} ${{ steps.versions.outputs.version_only }}"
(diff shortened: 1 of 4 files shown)


eeberhard:
> Seems potentially fragile to me (branch checking, adding more comments on an existing branch, using the same branch name instead of a unique one)

If using a unique branch name (for example, generated from current date and time) addresses most of those concerns, then I think that would be a worthwhile change to make

eeberhard (approved):
Thanks for the update. Based on the workflow test runs I think it's good to go.

For the previous v1-1-0 change that didn't get generated because of the failed workflow, could you please manually create a PR with the new static files? You can re-use the result of one of the test branches, i.e. https://github.com/aica-technology/api/pull/109/files, and just drop the other files / change the version back.

LouisBrunner (approved):

domire8 (approved):
Cool!

Issue #99, opened 2024-01-15
Add `required` field for empty parameters


domire8:
https://github.com/aica-technology/api/blob/c94ebae55013754c2533d551633abe1588c719b8/schemas/parameters/schema/parameter.schema.json#L1

We have components and controllers that have parameters which don't have a default value but need to be set such that the controller/component can run correctly. Is that something we should add to the schema such that we can visualize it in the frontend? 

What happened to me is that I added a timer and tried to run it without setting the `timeout` parameter (which by the way obivously shows `default` next to its empty value - which is an additional point of confusion maybe) and then I wondered why its not running.

I know it's a detail and quite hard to be strict about it because parameters can have no default value and not be required because the component handles that internally but they can also have empty parameters where we don't know what a sensible value is and then the user needs to set it.

Long story short, I think we can improve the experience here, either by tackling it
- through a more detailed schema
- or by having stricter parameter guidelines on the controller/component side where we say parameters cannot be empty if they are required to run the controller/component.

What are your thoughts @eeberhard ?

eeberhard:
The design intent was for `default: null` to indicate that a value is required. From [the docs](https://docs.aica.tech/docs/reference/custom-components/component-descriptions#parameters):

> When a parameter has no valid default state and must be set by the user for the component to function, the default_value property in the component description should be set to null. This is distinct from a valid default empty parameter state, which can be expressed with the empty string "".

But I agree it's not immediately clear, and certainly the UI doesn't distinguish this.

The reasoning is that a parameter that doesn't have a default value _must_ be set, therefore is required. Do you think there are any reasonable cases where you would say a parameter is _required_ even though it has default value? What about a parameter that doesn't have a default value but is also optional?

The problem with two fields is that we have four possible combinations; required with default value, required with no default, not required with default value, not required with no default value. Of those I think only "required with no default, not required with default value" make sense, in which case there is no point to have a `required` field.

For me it seems more like a documentation and UI issue to make required parameters more obvious. If there is no default value, we can highlight that parameter in the UI (instead of just showing it as the current "default")

domire8:
> Do you think there are any reasonable cases where you would say a parameter is required even though it has default value? 

I think no, that would be bad design

> What about a parameter that doesn't have a default value but is also optional?

That can happen, for example the reference pose and angular limit in the IK controller. If not set, we don't do anything with it.

I think the 'not required with no default value' is therefore also a valid case but no strong feelings..

eeberhard:
[...]

PR #111, opened 2024-01-16
feat(parameter-schema): add new keyword optional for the parameter schema

domire8:
## Description

- #99 

See parent issue for the discussion. The changes a schema are: `If the default value of a parameter is set to anything different from null, the property optional is prohibited`

## Review guidelines

## Checklist before merging:

### DIFF
--- a/schemas/parameters/schema/parameter.schema.json
+++ b/schemas/parameters/schema/parameter.schema.json
@@ -3,6 +3,7 @@
   "title": "Parameter",
   "description": "A dynamic parameter that is publicly configurable.",
   "type": "object",
+  "additionalProperties": false,
   "properties": {
     "display_name": {
       "description": "The human-readable parameter name.",
@@ -48,6 +49,10 @@
         }
       ]
     },
+    "optional": {
+      "description": "Specify if this parameter if optional if the default value is null",
+      "type": "boolean"
+    },
     "dynamic": {
       "description": "Specify if this parameter can be dynamically reconfigured.",
       "type": "boolean"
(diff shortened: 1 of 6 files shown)


jjenscodee (approved):

schemas/parameters/schema/parameter.schema.json
```diff
@@ -65,16 +70,39 @@
...
+          }
+        }
+      },
+      "then": {
+        "not": {
+          "required": [
+            "optional"
+          ]
+        }
+      }
```
eeberhard, line 90: I'm not certain about the logical implications here; does "not: required" equate to "forbidden"? if you've tested it with VSCode parser and it does what you expect, then that's great. But in my head it makes more sense if you phrase it the other way around: don't include the `optional` property to begin with, but if default value is null, then include `optional`and make it required.

```suggestion
      "if": {
        "properties": {
          "default_value": {
            "const": null
          }
        }
      },
      "then": {
        "properties": {
          "optional": {
            "description": "Specify if this parameter if optional if the default value is null",
            "type": "boolean"
          },
          "required": [
            "optional"
          ]
        }
      }
```

domire8: As explained in person, I believe that `not: required` actually means forbidden, especially with the VSCode parser.

schemas/parameters/schema/parameter.schema.json
```diff
@@ -48,6 +49,10 @@
         }
       ]
     },
+    "optional": {
+      "description": "Specify if this parameter if optional if the default value is null",
```
eeberhard, line 53: nitpick for wording and clarity
```suggestion
      "description": "Specify if this parameter is optional (only applicable if the default value is null)",
```

eeberhard (approved):
Small nitpick, then looks good to me 👍 hope the workflow to generate new PR from this version change is successful!

Issue #112, opened 2024-01-17
Schema for conditions


domire8:
https://github.com/aica-technology/api/blob/a0cb3ddd3958bd409fee7ee4d70df2031eaa5a63/schemas/applications/schema/conditions.schema.json#L23

Something does not quite work for the condition schema. I think what's happening with the "oneOf" is that it discards the event property and the `["required": "events"]`. That's what I can say from my tests. I get errors saying that `events` is not allowed, but when I remove `additionalProperties` from the condition_object or the predicate_object, there is no error anymore. So I think we have to rework this a bit, and I tried but can't seem to fix it. Any input appreciated.

@eeberhard @jjenscodee

domire8:
Or maybe what is happening is that we merge two schemas (the one from the event and the condition schema) and both of these schemas have `additionalProperties: false` so they kind of prohibit each other.

eeberhard:
Thanks for making the issue. I was aware of it for quite a while but didn't get around to reporting the issue or looking into it deeply. This part of the schema is very complex because of the recursive nature of conditions. The schema parser used by the frontend and by VSCode (redhat yaml extension) both struggle with this. When I wrote the condition schema originally, I used the parser in CLion which had no issues with the syntax. Ultimately it is probably an issue with the json-schema draft version.

I will have a look at alternative ways of describing the condition syntax that work with all parsers.

eeberhard:
While we are reworking the condition schema, this would be a good time to make `events` property optional and not required. It might even resolve the core issue, but moreover it allows conditions to be defined without them actually triggering events. This can be useful when using conditions through the API or sequences just as a "wait" condition, rather than an event-triggering condition.

domire8:
Yeah, if it wasn't recursive, we could just move the `event` property to the `condition_object`. Agree about making it optional

jjenscodee:
btw, I check the backend/examples/components/test_binary_recorder.yaml because it has the condition and I get lots of errors when opening it. Not sure if it's because this file hasn't been updated for a while? (p.s. I'm linking to application.schema.json locally, not from docs.aica.tech)

Do you have the same result? or the way I test is wrong/not related?
```
Property load_controller is not allowed.
Property interface is not allowed.
Property start is not allowed.
Incorrect type. Expected "Controller Identifier".
Property all is not allowed.
Property events is not allowed.
Incorrect type. Expected "array | Controller Identifier".
```
[...]

PR #114, opened 2024-01-22
fix: update conditions schema

domire8:
## Description

- #112 

This PR solves the issue by tweaking the conditions schema such that the parser (and the user) know which properties are allowed on the top level, and then the recursion only starts one level deeper. It is a bit of code duplication but still quite an elegant solution IMO. Tested especially in the DSE, so I'm quite confident this is good to go.

## Review guidelines

## Checklist before merging:


eeberhard (approved):
Great work with this!

jjenscodee (approved):

Issue #127, opened 2024-02-18
Define a syntax for sequenced events in the application schema


eeberhard:
This issue serves to improve support for sequential behavior with a new sequencing syntax. Sequential behavior means events that are explicitly triggered one after the other, rather than driven purely by component predicates.

In brief, a sequence is a list of steps that are triggered one after the other. Each sequence step can be a standard state event (load, unload, set parameter, etc). Sequence steps also include the option to wait; wait for a single predicate, wait for a condition, or wait for some timed duration.

After sequences are defined in the YAML, they can be started, restarted or aborted with a new `sequence` state events; component predicates, conditions and even sequences can also start, restart or abort a sequence. This approach is fully compatible with our event-driven and asynchronous behavior, but allows us to be more expressive with particular behaviors and sequential actions.

### Sequences explained

A simple sequence might look like this:

```yaml
sequences:
  my_sequence:
    - load:
        component: foo
    - wait:
        component: foo
        predicate: is_active
    - service:
        component: foo
        name: do_something
    - wait:
        seconds: 5
    - set:
        component: foo
        parameter: bar
        value: baz
```

The `wait` steps for component predicates or conditions can have an optional  timeout block, which define breakout events in case the predicate or condition are not true within the time limit

```yaml
wait:
  condition: foo
  timeout:
    seconds: 10.0
    events:
      unload:
        component: foo
```

There is also an `assert` which checks if a predicate or conditions is true right now, and otherwise triggers some breakout events

```yaml
assert:
  component: foo
  predicate: bar
  else:
    unload:
      component: foo
```

In both cases when breakout events are triggered, it aborts the current sequence. The syntax can maybe still be improved, but overall it allows for error recovery and branching logic (including sequences that start other sequences conditionally)

Finally, to manage sequences, the `sequence` event is used with either `start`, `restart`, or `abort` pointing to the name of the sequence:

```yaml
on_start:
  sequence:
    start: my_sequence
```

PR #128, opened 2024-02-18
feat(schemas): sequences

eeberhard:
## Description

- #127 

This PR addresses the issue by defining a sequencing syntax in the application schema.

To validate mutually exclusive fields in `wait` and `assert`, I used a trick with `oneOf`. Because `oneOf` must always validate one and only one subschema, putting `{}` as the first item under `oneOf` always validates first and effectively prohibits all subsequent schemas.

## Review guidelines

Estimated Time of Review: 20 minutes

## Checklist before merging:

### DIFF
--- a/schemas/applications/CHANGELOG.md
+++ b/schemas/applications/CHANGELOG.md
@@ -2,10 +2,17 @@
 
 Release Versions:
 
+- [1-2-0](#120)
 - [1-1-2](#112)
 - [1-1-1](#111)
 - [1-1-0](#110)
 - [1-0-0](#100)
+
+## 1-2-0
+
+Version 2-0-0 adds a new syntax to manage sequential events as an array of steps to be handled in order. Sequence steps
+are either standard state events or conditional blocks; the latter are used either to wait for a condition, predicate
+or fixed time interval, or to assert the current value of a condition or predicate.
 
 ## 1-1-2
 
--- a/schemas/applications/schema/application.schema.json
+++ b/schemas/applications/schema/application.schema.json
@@ -24,6 +24,9 @@
         "switch_controllers": {
           "$ref": "events/switch_controllers.schema.json"
         },
+        "sequence": {
+          "$ref": "events/sequence.schema.json"
+        },
         "position": {
           "$ref": "common/position.schema.json"
         }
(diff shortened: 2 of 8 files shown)


eeberhard:
I used the following test cases to check validation behavior, which I think are mostly exhaustive. All items passed / failed as expected.

```
sequences:
  valid_sequence:
    - load:
        component: foo
    - wait:
        seconds: 10
    - wait:
        component: foo
        predicate: foo
    - wait:
        condition: foo
    - wait:
        component: foo
        predicate: foo
        timeout:
          seconds: 10
          events:
            load:
              component: foo
            unload:
              component: foo
    - wait:
        condition: foo
        timeout:
          seconds: 10
          events:
            load:
              component: foo
            unload:
              component: foo
    - assert:
        component: foo
        predicate: foo
    - assert:
        condition: foo
    - assert:
        component: foo
        predicate: foo
        else:
          load:
            component: foo
          unload:
            component: foo
    - assert:
        condition: foo
        else:
          load:
            component: foo
          unload:
            component: foo
  invalid_sequence:
    - wait:
        seconds: 10
        timeout: { }
    - wait:
        seconds: 10
        timeout:
          seconds: 10
          events: { }
    - wait:
        component: foo
    - wait:
        predicate: foo
    - wait:
        condition: foo
        predicate: foo
    - wait:
        condition: foo
        component: foo
    - wait:
        component: foo
        predicate: foo
        seconds: 10
    - wait:
        condition: foo
        seconds: 10
    - wait:
        condition: foo
        timeout: { }
    - wait:
        condition: foo
        timeout:
          seconds: 10
    - wait:
        condition: foo
        timeout:
          events: { }
    - assert:
        component: foo
    - assert:
        predicate: foo
    - assert:
        condition: foo
        component: foo
    - assert:
        condition: foo
        predicate: foo
    - assert:
        else: {}
```

domire8 (approved):
Just two small corrections to make.

As [this thread on stackoverflow](https://stackoverflow.com/questions/28162509/mutually-exclusive-property-groups) mentions, the `oneOf: {}` trick is slightly less readable but probably the right way to go. I was hoping we could do something with `not: required: ["field"]` but it doesn't work easily.

One thing that I noticed was that the YAML interpreter in VSCode is really confused by this syntax and while it's showing the errors correctly, the "error messages" or "hints" are useless, i.e. they give the wrong information about what should be fixed

domire8 (approved):
Indeed a bit more readable! Thanks

Issue #139, opened 2024-03-22
Update docs for signal names


domire8:
Following https://github.com/aica-technology/modulo/issues/66, the goal is to add a note in the docs that topic names should be `lower_snake_case`. We don't have to wait for a release since this is already true now.

PR #141, opened 2024-03-22
feat: add note about signal name conventions

domire8:
## Description

- #139 

This PR solves the issue by adding a caution block to the relevant section

## Review guidelines

## Checklist before merging:

### DIFF
--- a/docs/docs/reference/custom-components/02-custom-component.md
+++ b/docs/docs/reference/custom-components/02-custom-component.md
@@ -325,7 +325,7 @@ be called `foo_topic`.
 
 For these reasons, signal names must adhere to the following rules:
 
-- Signal names must be unique across all inputs and outputs
+- Signal names must be unique for all inputs and outputs of the component
 - Signal names must be written in `lower_snake_case` (using only lowercase letters, numbers and underscores)
 - Signal names cannot start with a number or underscore.
 
--- a/docs/docs/reference/custom-components/02-custom-component.md
+++ b/docs/docs/reference/custom-components/02-custom-component.md
@@ -315,6 +315,20 @@ The binding logic is supported for a number of common message data types. To int
 (for example, to communicate with ROS 2 nodes using custom message types outside the AICA framework), it is possible to
 define raw publishers and subscribers following standard ROS 2 conventions.
 
+### Signal name
+
+Component signals are declared and managed with unique signal names. The signal name is used to determine the default
+topic. For a signal added under the name `foo`, the default topic will be `~/foo`. Adding a signal also automatically
+creates and associates a parameter with that signal with the parameter name `[signal_name]_topic`, which is used to
+override the default topic name at runtime. For a signal name `foo`, the parameter associated to the signal topic will
+be called `foo_topic`.
+
+For these reasons, signal names must adhere to the following rules:
+
+- Signal names must be unique across all inputs and outputs
+- Signal names must be written in `lower_snake_case` (using only lowercase letters, numbers and underscores)
+- Signal names cannot start with a number or underscore.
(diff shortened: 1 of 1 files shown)


bpapaspyros (approved):

docs/docs/reference/custom-components/02-custom-component.md
```diff
@@ -554,6 +554,14 @@ Lifecycle components publish outputs only when the component is in the `ACTIVE`
 
 :::
 
+:::caution
+
+Each signal shall be named using only characters from the alphanumeric set (letters A-Z and numbers 0-9) and the
+underscore (`_`). Additionally, the `lower_snake_case` convention should be used and signal names may not start with a
+number or an underscore. 
+
+:::
```
eeberhard, line 563: I would not bury this down at the end of the signals section. You also use capitals when you write `alphanumeric set (letters A-Z and numbers 0-9)`, but it would be preferable to document only `lower_snake_case` . I would rather structure and rewrite it like this:


```markdown
## Adding signals
[no change]

### Signal name

Component signals are declared and managed with unique signal names. The signal name is used to determine the default
topic. For a signal added under the name `foo`, the default topic will be `~/foo`. Adding a signal also automatically
creates and associates a parameter with that signal with the parameter name `[signal_name]_topic`, which is used to
override the default topic name at runtime. For a signal name `foo`, the parameter associated to the signal topic will
be called `foo_topic`.

For these reasons, signal names must adhere to the following rules:

- Signal names must be unique across all inputs and outputs
- Signal names must be written in `lower_snake_case` (using only lowercase letters, numbers and underscores)
- Signal names cannot start with a number or underscore.

## Message types
[no change]
```

eeberhard (commented):
You should cross reference from the component-descriptions.md page to the #signal-name section in the custom-component.md page, to explicitly say that the signal name in the description must match the declared signal name in the implementation. Ideally we would repeat this for parameters and all other interfaces to be explicit.

docs/docs/reference/custom-components/02-custom-component.md
```diff
@@ -315,6 +315,20 @@ The binding logic is supported for a number of common message data types. To int
...
+
+Component signals are declared and managed with unique signal names. The signal name is used to determine the default
+topic. For a signal added under the name `foo`, the default topic will be `~/foo`. Adding a signal also automatically
+creates and associates a parameter with that signal with the parameter name `[signal_name]_topic`, which is used to
+override the default topic name at runtime. For a signal name `foo`, the parameter associated to the signal topic will
+be called `foo_topic`.
+
+For these reasons, signal names must adhere to the following rules:
+
+- Signal names must be unique across all inputs and outputs
```
domire8, line 328:
```suggestion
- Signal names must be unique across all inputs and outputs of a component
```

domire8:
> to explicitly say that the signal name in the description must match the declared signal name in the implementation. 

Isn't that already the case? https://docs.aica.tech/docs/reference/custom-components/component-descriptions#signals

docs/docs/reference/custom-components/02-custom-component.md
eeberhard: Good point, maybe even more clear as:
```suggestion
- Signal names must be unique for all inputs and outputs of the component
```

eeberhard (approved):

eeberhard:
> > to explicitly say that the signal name in the description must match the declared signal name in the implementation.
> 
> Isn't that already the case? https://docs.aica.tech/docs/reference/custom-components/component-descriptions#signals

Yes, I was looking for an actual link, but the writing is clear enough as is.

PR #148, opened 2024-04-19
fix: application schema 1-4-0 and controller schema 1-3-0

domire8:
## Description

- https://github.com/aica-technology/api/actions/runs/8752660355

The one time I don't test my changes..

## Review guidelines

Estimated Time of Review: 1.1 minutes

## Checklist before merging:

### DIFF
--- a/docs/static/schemas/1-3-0/controller.schema.json
+++ b/docs/static/schemas/1-3-0/controller.schema.json
@@ -1,0 +1,419 @@
+{
+  "$id": "/1-3-0/controller.schema.json",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "AICA Controller Description",
+  "description": "A full description of a controller in the AICA application framework.",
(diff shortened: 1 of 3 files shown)


eeberhard (commented):
Incorrect formatting of brackets was hard to spot in the diff, understandable.

But I disagree with increasing the patch version here. Since the previous attempt 1-4-0 was not a valid schema (no static files generated), it means 1-4-0 was not actually released, and 1-4-1 therefore makes no sense. Rather, I would run and deploy the bundled schema for 1-4-0 manually in this case.

domire8:
Good call, but it doesnt have a manual dispatch trigger..

eeberhard (commented):
> Okay and now I realize you probably meant to generate it locally manually and add it to this PR?

yes 👍 but you also need to push the controller description 1-3-0 which was part of #146

eeberhard (approved):

PR #151, opened 2024-04-22
fix: invalid nesting of required properties in schema 1-4-0

eeberhard:
## Description

After #146 introduced controller events to the application schema with a missing closing bracket, #148 "fixed" this problem but added the closing bracket in the wrong place. As a result, the required properties were nested _under_ the controller and hardware properties, rather than being next to them.

Overly simplified view of how the incorrect nesting for required properties looked after #148:
```
hardware: {
  properties: {
    controllers: {
      properties: {
        events: {
          ...
        },
        required: [ ],
      },
    },
    required: [ ]
  }
}
```

Overly simplified view of how the correct nesting for required properties should look (fixed in this PR):
```
hardware: {
  properties: {
    controllers: {
      properties: {
        events: {
          ...
        }
      },
      required: [ ],
    }
  },
  required: [ ]
}
```

Since the previous deployment of schema 1-4-0 failed to validate existing applications, this PR revokes and replaces the statically generated schema with the correctly intended version of 1-4-0.

## Review guidelines

## Checklist before merging:

### DIFF
--- a/docs/static/schemas/1-4-0/application.schema.json
+++ b/docs/static/schemas/1-4-0/application.schema.json
@@ -450,19 +450,19 @@
                           "$ref": "#/properties/sequences/$defs/timeout/properties/events"
                         }
                       }
-                    },
-                    "required": [
-                      "plugin"
-                    ]
-                  }
+                    }
+                  },
+                  "required": [
+                    "plugin"
+                  ]
                 }
               }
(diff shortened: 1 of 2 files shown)


domire8 (approved):
Oh sorry about that, thanks for catching it!

Issue #152, opened 2024-04-24
Support controller predicates in conditions and sequences


domire8:
As title suggests.

PR #153, opened 2024-04-24
feat: support controller predicates in conditions and sequences

domire8:
## Description

- #152 

This PR solves the issue by extending the schema once again. At this point we have to think if there is something we can do to reduce code duplication, e.g. define a predicate object that has the controller, hardware, component properties with the `oneOf` `required`. Any suggestions?

## Review guidelines

## Checklist before merging:

- [ ] Update docs and examples

### DIFF
--- a/docs/docs/reference/02-yaml-syntax.md
+++ b/docs/docs/reference/02-yaml-syntax.md
@@ -914,7 +914,7 @@ my_button:
 
 ## Validating a YAML application
 
-The [YAML application schema](https://docs.aica.tech/schemas/1-4-0/application.schema.json) defines the structural rules
+The [YAML application schema](https://docs.aica.tech/schemas/1-5-0/application.schema.json) defines the structural rules
 of an AICA application and effectively distinguishes between valid and invalid syntax.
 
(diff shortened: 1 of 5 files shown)


eeberhard (commented):
Looks alright to me, despite the duplication. We could probably define an "application predicate" object which is one of component or controller. Either way I'll probably refactor the definitions for model 2-0-0 to define predicates, conditions and states consistently everywhere, so I think it's fine for now.

This is just an addition and extension to the previous version, please use 1-4-1 and not 1-5-0. See this design document as reference: https://docs.snowplow.io/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver/

I haven't been very strict with it in the past (including 1-2-0, 1-3-0 and 1-4-0) but this is certainly a case for an addition to the previous version.

docs/docs/reference/02-yaml-syntax.md
```diff
@@ -621,8 +637,15 @@ conditions:
     predicate: ...
     events:
       ...
-
+
```
domire8, line 640:
```suggestion

```

eeberhard (approved):

Issue #156, opened 2024-05-02
`wait_for` return dict


domire8:
https://github.com/aica-technology/api/blob/1416d892dfdab0d2ceeb6a5b8b7733b6a13dda7d/python/src/aica_api/client.py#L279

@eeberhard since #95, the python client `wait_for...` functions return `Union[None,dict]` instead of bool. Is this an oversight or is there a reason for it?

@yrh012

eeberhard:
The changes to `read_until` were intentional, but the return type of the calling `wait_for` functions was not intended to change. The `wait_for` functions should still return True or False depending on if the waiting ended before the timeout or not.

The reason it wasn't caught is that `None` and a wellformed `dict` are respectively falsey and truthy when evaluated in a boolean condition such as:
```python
if wait_for_component(...):
    # do something if true
else:
    # do something if false
```

This can be made more explicitly into a boolean return type using `return read_until(...) is not None`.

eeberhard:
I will also use this issue as a springboard to add `wait_for_hardware`, `wait_for_controller` and `wait_for_controller_predicate`, since they are not currently represented.

PR #157, opened 2024-05-05
feat: support hardware and controller states and predicates in `wait_for_` functions

eeberhard:
## Description

- #156 

This PR solves the issue by fixing the return type of the `wait_for_` functions using an explicit `is not None` operator.

It also adds new functions to wait for controller and hardware states and controller predicates. As part of this, the `wait_for_predicate` has been marked as deprecated in favor of the equivalent `wait_for_component_predicate`, in order to to disambiguate component and controller predicates more effectively.

## Review guidelines

## Checklist before merging:

### DIFF
--- a/python/CHANGELOG.md
+++ b/python/CHANGELOG.md
@@ -10,6 +10,10 @@ Release Versions:
 - [1.0.1](#101)
 - [1.0.0](#100)
 
+## Upcoming changes
+
+- feat: support hardware and controller states and predicates in `wait_for` functions (#156)
+
 ## 2.0.0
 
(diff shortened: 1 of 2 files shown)


domire8 (approved):
Nice, thanks! Do you want to release that?

yrh012 (approved):
Nice!

eeberhard:
I will merge this and release it as a separate PR. As part of the release I want to make a note that hardware data is only available from API version 3.1 onwards. While this isn't a breaking change, I still want to capture it in the `check()` function and the compatibility table in the README.

I've captured that followup issue here:
- https://github.com/aica-technology/api/issues/158

Issue #158, opened 2024-05-08
Define more detailed feature and function compatibility between versions


eeberhard:
Minor updates to the core API version (for example, from v3.0 to v3.1) introduce new endpoints and features. The Python client can correspondingly add new wrapper functions for these endpoints or features, equally released under minor updates.

In order to know which features are available with which versions, the compatibility table and `check()` function should have corresponding indications. Additionally, it would be nice for a function decorator to apply a specific version check and warn if a feature is not available in the current version.

This is different to breaking changes which are marked by major changes to the API version or Python client. These normally mean that the code is no longer backwards compatible (things that worked before don't work anymore).

PR #159, opened 2024-05-09
feat(client): define more detailed feature and function compatibility between versions

eeberhard:
## Description

- #158 

Check feature and function version compatibility more precisely:

* Use deprecation module to warn deprecated function with version range
* Add helper and local class attribute to get and store api server version
* Define decorator to require a minimum api server version when calling a function, otherwise warn and return None
* Update check() function with more cases
* Use semver module to compare and match versions instead of string comparisons
* Use logging module instead of printing warnings
* build: add and revise dependencies

## Supporting information

To test this out properly (since I changed some dependencies), I pushed a release candidate to the test registry at https://test.pypi.org/project/aica-api/2.1.0rc5. As part of this, I learned that doing a pip install from this registry will also try to install all dependencies from the test registry. This will fail against the version checks, since not every package and certainly not every latest version is on the test registry. In this particular case, if you want to test the prebuilt version from the test registry, you need to install or upgrade all dependencies manually first:
```
## first install dependencies manually from the standard pypi registry
python3 -m pip install --upgrade pyyaml deprecation semver python-socketio
## then install the release candidate from the test.pypi registry
python3 -m pip install -i https://test.pypi.org/simple/ aica-api==2.1.0rc5
```
Another option for testing of course is to install locally from this source repo (`pip install .`).

In general, I think it's preferable to make actual release candidate builds for the official registry (using the pre-release specifier `-rc.X`), which would be built and pushed here automatically after merging. Pre-release builds are not listed or installed by default when just using `pip install aica-client`, since it will always use the latest stable version in that case. If we see all behaviors and dependencies work as expected after testing, we can officially release by removing the `-rc.X` specifier and updating the changelog. 

## Review guidelines

I checked it with a AICA `base` images v3.1 and v3.2, which include backend-API server versions v3.0.3 and v3.1.0 respectively. It correctly warned for the unsupported functions (wait for controller, hardware, etc).

## Checklist before merging:

### DIFF
--- a/python/CHANGELOG.md
+++ b/python/CHANGELOG.md
@@ -13,6 +13,7 @@ Release Versions:
 ## Upcoming changes
 
 - feat: support hardware and controller states and predicates in `wait_for` functions (#156)
+- feat: define more detailed feature and function compatibility between versions (#158)
 
 ## 2.0.0
 
--- a/python/README.md
+++ b/python/README.md
@@ -42,15 +42,21 @@ else:
 
 ## Compatability table
 
-The latest version of the AICA API client will generally support the latest API version in the AICA framework. For
-older versions of the AICA framework, it may be necessary to install older versions of the client. Use the following
-compatability table to determine which client version to use.
+The latest version of the AICA API client will generally support the latest API server version in the AICA framework.
+Major changes to the API client or server versions indicate breaking changes and are not backwards compatible. To
+interact with older versions of the AICA framework, it may be necessary to install older versions of the client.
+Use the following compatability table to determine which client version to use.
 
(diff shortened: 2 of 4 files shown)


domire8 (approved):
Neat solution with the decorator!

domire8:
When will we then release the actual version?

eeberhard:
> When will we then release the actual version?

I would merge this which would build and push the 2.1.0 release candidate, and then I would make a separate little release PR for 2.1.0 right after testing. Mainly I just want to make sure all dependencies work as expected and the release candidate is a nice way of testing that through pypi.

LouisBrunner (approved):

Issue #147, opened 2024-04-16
Draft application syntax 2.0


eeberhard:
# Scope

This issue is to propose the next version of the AICA application syntax (version 2.0).

The goal is to define all the desired changes to the application syntax, including any breaking changes, to enable even better composition of applications. All proposals are up for discussion under this issue, including any syntax choices made in the examples.

Many of the suggested schema revisions would also require changes to modulo components, controllers, the Dynamic State Engine, and the implementation of the Developer Interface. This proposal does not go into details, but consider that some changes have more associated consequences in implementation work than others.

## Outline of proposals

This proposal is structured in a certain way to group related concepts. This forms a narrative outline:

- [A holistic approach to lifecycle states](#a-holistic-approach-to-lifecycle-states)
- [Automatic behaviors (auto-events)](#automatic-behaviors)
- [Restrict names to lower_snake_case](#restrict-all-names-and-namespaces-to-lower_snake_case)
- [Positions](#positional-data-in-a-separate-field)
- [Display names](#display-names)
- [Application events](#application-events)
- [Static frames](#static-frames)
- [Component frames](#components-as-consumers-and-producers-of-frames)
- [Metadata](#metadata)

## Types of changes

The proposal could also be structured by the type of change. I have used the following labels throughout the proposal:

> ⚠️ _**BREAKING CHANGE**_ ⚠️

Some parts of the proposal constitute breaking changes in the syntactical sense (old applications will not validate against the new schema).

> 🤔 _CONCEPTUAL CHANGE_ 🤔

Some changes are conceptual changes even if not technically breaking (old applications may still validate against the new schema, but may not behave as expected).

> 💡 _ENHANCEMENT_ 💡

Some changes are just ideas for enhancements that could be an eventual revision or addition to the schema.

# Proposals

## A holistic approach to lifecycle states

A main feature of an AICA application is that the states of all components and controllers can change dynamically (ergo, the name Dynamic State Engine). Since we already track the states (loaded or unloaded, including lifecycle states) of all elements, we could make better use of these states in conditional logic, and use the transitions of states to drive events.

### State transitions as event drivers

Currently, we broadcast component lifecycle states through predicates (`is_unconfigured`, `is_inactive`, etc). We also broadcast a fake `on_load` predicate with every loaded component that is set true when the component is constructed, and automatically set false thereafter.

[...]

Issue #140, opened 2024-03-22
Restrict names in schema to lower_snake_case (breaking change)


domire8:
> As a small note, our application schema restricts most namespaces to the regexp `(^[a-zA-Z][a-zA-Z0-9_]*[a-zA-Z0-9]$)|(^[a-zA-Z]$)`, so it does not actually enforce lower case for things like named service calls in the YAML. Changing it to `(^[a-z][a-z0-9_]*[a-z0-9]$)|(^[a-z]$)` would be nice but would be a breaking change to the schema unfortunately (potentially invalidates previously valid schemas).

> Still, we should suggest `lower_snake_case` in the docs and automatically sanitize to lower case in the implementation, and update the schema when the next breaking change happens.

_Originally posted by @eeberhard in https://github.com/aica-technology/modulo/issues/66#issuecomment-2012795000_

PR #162, opened 2024-06-24
feat(schema): application syntax draft/2-0-0

eeberhard:
## Description

- #147 

## Review guidelines

Estimated Time of Review: 90 minutes

Review by commit

## Checklist before merging:


eeberhard:
Example of the new syntax:

```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/aica-technology/api/feat/draft-v2.0.0/docs/static/schemas/draft/2-0-0/application.schema.json

schema: draft/2-0-0

dependencies:
  base: v4.0.0
  packages:
    - { package: "@aica/components/pose-detection", version: "v1.0.0" }
    - { package: "@aica/collections/ur-collection", version: "v3.1.0", constraints: ">= 3.0.0" }

metadata:
  name: My application
  description: Some description of the application
  tags:
    - "motion"
    - "force control"
    - "universal robots"

userdata:
  info: "..."
  website: github.com/eeberhard

on_start:
  load:
    - component: my_component
    - hardware: my_hardware
  sequence:
    start: my_sequence

conditions:
  my_condition:
    display_name: My cool condition
    condition:
      any:
        - component: foo
          state: active
        - controller: foo
          hardware: bar
          predicate: baz
    events:
      unload:
        component: foo

sequences:
  my_sequence:
    display_name: My cool sequence
    steps:
      - delay: 10
      - check:
          condition:
            condition: my_condition
          else:
            load:
              component: foo
      - check:
          condition:
            component: foo
            state: active
          wait_forever: true
      - check:
          condition:
            sequence: foo
            state: active
          timeout: 10
          else:
            unload:
              component: foo

hardware:
  my_hardware:
    urdf: foo
    rate: 12.5    
    rate_tolerance: 0.99
    strict: true
    events:
      transitions:
        on_load:
          load:
            controller: my_controller
            hardware: my_hardware
    controllers:
      my_controller:
        plugin: my_controller/Foo
        events:
          transitions:
            on_load:
              switch_controllers:
                hardware: my_hardware
                activate: my_controller
          predicates:
            is_stable:
              sequence:
                start: my_sequence

components:
  foo:
    component: my_package::Foo
    parameters:
      a: foo
      b: bar
    inputs:
      my_input: /foo/my_input
    outputs:
      my_output: /foo/my_output
    events:
      transitions:
        on_configure:
          lifecycle: activate
      predicates:
        is_in_range:
          unload:
            component: foo

graph:
  positions:
    on_start:
      x: 20
      y: 40
    conditions:
      my_condition:
        x: 0
        y: 0
    sequences:
      my_sequence:
        x: 0
        y: 0
    components:
      foo:
        x: 0
        y: 0
    hardware:
      my_hardware:
        x: 0
        y: 0
```

docs/static/schemas/draft/2-0-0/application.schema.json
```diff
@@ -0,0 +1,1564 @@
...
+                    "log_level": {
+                      "$ref": "#/$defs/log_level"
+                    },
+                    "plugin": {
+                      "title": "Controller Plugin",
+                      "description": "The name of the controller plugin",
+                      "type": "string"
+                    },
+                    "parameters": {
+                      "$ref": "#/$defs/parameters"
```
domire8, line 282: https://github.com/aica-technology/dynamic-state-engine/issues/278

Do we "allow" the rate parameter for controllers under the `parameters` field like we do for components or do we need to add it as a separate property?

docs/static/schemas/draft/2-0-0/application.schema.json
```diff
@@ -0,0 +1,1564 @@
...
+                  "type": "object",
+                  "additionalProperties": false,
+                  "properties": {
+                    "on_load": {
+                      "$ref": "#/$defs/events"
+                    },
+                    "on_configure": {
+                      "$ref": "#/$defs/events"
+                    },
+                    "on_activate": {
```
domire8, line 435: Why do we go from `is_active` to `on_activate`?

docs/static/schemas/draft/2-0-0/application.schema.json
```diff
@@ -0,0 +1,1564 @@
...
+                },
+                "call_service": {
+                  "$ref": "#/$defs/events/properties/call_service"
+                },
+                "lifecycle": {
+                  "$ref": "#/$defs/events/properties/lifecycle"
+                },
+                "switch_controllers": {
+                  "$ref": "#/$defs/events/properties/switch_controllers"
+                }
```
domire8, line 509: Need the set parameter event here

docs/static/schemas/draft/2-0-0/application.schema.json
```diff
@@ -0,0 +1,1564 @@
...
+                    },
+                    "hardware": {
+                      "$ref": "#/$defs/events/properties/call_service/oneOf/2"
+                    },
+                    "state": {
+                      "title": "Controller State",
+                      "description": "The runtime state of the controller",
+                      "type": "string",
+                      "enum": [
+                        "unloaded",
```
domire8, line 663: Just to be sure, are those really all the states that the DSE can know about controllers? no inactive or configured?

domire8 (commented):
I reviewed mainly the generated schema file to get the high level understanding and I'm mostly happy!

1. It's a pity that now under `events` we need to have an additional property `transitions` or `predicates` since that increases the number of things you need to write and remember. No real solution to that because we need to differentiate between predicates and states...
2. I'm not sure I like the `transitions: on_load` syntax/choice of words in the hardware (I think it works better for components). To me, the `load` is not a transition and I think we should choose words that emphasize that whatever events you put under this block will happen **after** the load happened. This is because the `load` event of a hardware is blocking and not asynchronous (like it's the case for components). Just looking at 
```
hardware:
  my_hardware:
    urdf: foo
    rate: 12.5    
    rate_tolerance: 0.99
    strict: true
    events:
      transitions:
        on_load:
          load:
            controller: my_controller
            hardware: my_hardware
    controllers:
      my_controller:
        plugin: my_controller/Foo
        events:
          transitions:
            on_load:
              switch_controllers:
                hardware: my_hardware
                activate: my_controller
```
I'm thinking that you could run into race conditions here since it states that on load, we activate the controller but how can you be sure that the controller is already loaded.. It's very subtle and a bit hard to describe, we can discuss in person if it's not clear.

eeberhard:
Thanks for the feedback! Actually a lot of your points come down to this new way of triggering events from state transitions.

> now under events we need to have an additional property transitions or predicates 

As you said, we need to distinguish these two types of event drivers. I considered a flatter structure like this but I didn't like it as much:

```yaml
my_component:
  component: my_component::Foo
  transition_events:  # or just `transitions:`
    on_load: ...
  predicate_events:  # or just `predicates:`
    is_in_range: ...
```

> on load, we activate the controller but how can you be sure that the controller is already loaded..

Predicates trigger events on any rising edge. But the intention with state transition events is that they trigger an event iff the specific transition is completed. In the implementation of transition events, I would only trigger `on_activate` events after the component / controller / etc has actually been activated. Because the underlying engine already manages and monitors the states of all components, we can actually avoid race conditions with this pattern better than the current implementation.

I can see the confusion because in the current implementation, `on_load` is a fake predicate that is actually triggered asynchronously with component loading and does not actually guarantee that the component has been fully constructed. The other potentially confusing point is that a transition like `on_activate` looks a lot like the transition callback function in the modulo component SDK (which is partly intentional), but in the SDK that function is called _during_ the transition, not after. On that note...

> Why do we go from is_active to on_activate?

I wanted to use a recognizable and consistent "transition callback" language similar to the component SDK. The previous predicate formulation `is_active` triggered the event every time the component state changed from not active to active. For that particular transition it is equivalent. But now consider the `is_inactive` predicate, which could be triggered both after configuring or after deactivating. So the name of state transitions events should reflect the triggering transition like `on_configure` and `on_deactivate`.

You might call it `after_load`, `after_configure`, ... but to me the `on_` feels right.

> Are those really all the states [...] about controllers? no inactive or configured?

Actually controllers have full lifecycle states but are automatically configured and go to `inactive` when loaded. The only transitions we really control are loading, activating, deactivating and unloading (and thereby the only primary states the controller stays in are unloaded, inactive or active).

Part of the reason of this being a draft is to just try the implementation and see what states actually make sense to control and expose for hardware and controllers.

domire8:
> As you said, we need to distinguish these two types of event drivers. I considered a flatter structure like this but I didn't like it as much

Agreed.

> Because the underlying engine already manages and monitors the states of all components, we can actually avoid race conditions with this pattern better than the current implementation.

Absolutely, and this will constitute a huge improvement obviously. I know all that but that underlying mechanism might not fully transparent to the user.

> I wanted to use a recognizable and consistent "transition callback" language similar to the component SDK

I agree about similar language but part of me still think it's too similar with the `on_activate`. A user that writes a component with the SDK has to know what to put in the `on_activate_callback` and then we tell him that if you put `on_activate` into the YAML app, it will be triggered **after** on activate. I don't know...

`after_` just doesn't look nice as a prefix :smile: I think we start with this version and see how we feel about it later

domire8 (approved):
> Part of the reason of this being a draft is to just try the implementation and see what states actually make sense to control and expose for hardware and controllers.

:fire:

docs/static/schemas/draft/2-0-0/application.schema.json
eeberhard: good point. If `rate` is a separate field for hardware, and we want that to apply to controllers and components equally, then I think it makes sense to have that as a separate optional field in each case, rather than assuming it under parameters.

yrh012:
Very cool stuff! 🔥

One thing that we should consider is enforcing the jsonschema to not allow a list of load/lifecycle events in the schema.

domire8:
> One thing that we should consider is enforcing the jsonschema to not allow a list of load/lifecycle events in a given sequence step.

Can you elaborate with an example? Is that something that's already not correct in the current schema or did we miss this in this PR?

domire8:
This PR also addresses (and closes) issues
- https://github.com/aica-technology/api/issues/140
- https://github.com/aica-technology/dynamic-state-engine/issues/278
- https://github.com/aica-technology/dynamic-state-engine/issues/242

yrh012:
> > One thing that we should consider is enforcing the jsonschema to not allow a list of load/lifecycle events in a given sequence step.
> 
> Can you elaborate with an example? Is that something that's already not correct in the current schema or did we miss this in this PR?
Here is an example with a small sequence that loads a list of components. In the current implementation, this is allowed in the schema but fails at execution indicating invalid load payload.

  prepare_components:
    # load the components
    - load: 
        - component: assembled_frame_broadcaster
        - component: approach_frame_broadcaster
        - component: approach_tf_to_signal

eeberhard:
> This PR also addresses issues [...]

Thanks @domire8 !

> One thing that we should consider is enforcing the jsonschema to not allow a list of load/lifecycle events in a given sequence step

I think this is something to discuss further in the parent issue #147, but as @yrh012 points out, the current schema allows a sequence step to contain a "multi-event structure". However, in the current backend implementation (using base v3.4.0 for example), a sequence step can only support a single event.

This basically means there is a discrepancy between the current schema and implementation, and the observation of that is more of a general bug / issue report. However the relevant thing to discuss for the point of syntax 2.0 is whether to [1] change the schema to match the implementation (restrict a sequence step to a single event) or [2] do not change the schema and instead fix the implementation to support a multi-event structure on a sequence step. The option [1] is a breaking syntax change and so would need t be part of 2.0.

eeberhard:
As discussed, I will merge this now such that we can start to roll out the draft/2-0-0 implementations. We can then make additional changes based on the comments as discussed before we officially release it.

Issue #167, opened 2024-08-26
Remove self-targeting events for components (services, parameters, lifecycle)


eeberhard:
The first version of the application syntax allows component predicate-driven events to be self-targeting; the target component did not need to be specified in the event object. For example:

The following syntax would explicitly target component `foo` for these events:

```yaml
lifecycle:
  component: foo
  transition: activate
call_service:
  component: foo
  service: foo
set:
  component: foo
  parameter: foo
  value: bar
```

But if a predicate on component `foo` was calling these events, then the component target could be omitted.

```yaml
lifecycle: activate
call_service:
  service: foo
set:
  parameter: foo
  value: bar
```

The problem is that controllers also have services and dynamically settable parameters. In addition, these events can be triggered by other sources, such as conditions and sequences. This means that, while the syntax would permit a formulation like `lifecycle: activate` as valid, it would fail to function as expected when included under a controller, condition or sequence event.

The simple proposal therefore is to forbid this "self-targeting" syntactical shortcut. As we move more and more towards authoring our applications through the UI and not by hand, verbose but explicit syntax should be preferred to ambiguous shorthands. The plus side to being more restrictive with syntax is that it also simplifies the parser implementations.

PR #171, opened 2024-08-26
feat(schema): remove self targeting events

eeberhard:
## Description

- #167 

Here, I enforce lifecycle, parameter and service events to include a target component as per the issue description. I also removed a legacy "string" type structure for call service which used to allow `call_service: foo` instead of `call_service: { service: foo }`.

I already updated the changes to the generated schemas in the docs/static directory as part of this PR, since the version hasn't changed. Normally we are not allowed to manually update the static deployments for backwards compatibility reasons, but here the `draft` path is intended to allow us to make these breaking changes for downstream testing before the official release.

## Review guidelines

## Checklist before merging:

### DIFF
--- a/docs/static/schemas/draft/2-0-0/application.d.ts
+++ b/docs/static/schemas/draft/2-0-0/application.d.ts
@@ -47,10 +47,6 @@ export type LoadOnStart = ComponentIdentifier | HardwareIdentifier;
  */
 export type URDF = string;
 /**
- * The rate in Hz at which to run the hardware interface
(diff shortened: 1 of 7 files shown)


domire8 (approved):

jjenscodee:
Hello,

I got this error from the parser in frontend:

Can't resolve reference #/properties/on_start/$defs/load_on_start/oneOf/1/properties/hardware from id /draft/2-0-0/application.schema.json

PR #172, opened 2024-09-02
fix(schema): define a property at top-level for simpler bundling

eeberhard:
## Description

Reference the identifier.schema.json in the top-level application schema so that the schema bundler uses this high-level reference in place of the first occurence (which was previously buried under a complex path)

Thanks to @jjenscodee for noticing an issue here (https://github.com/aica-technology/api/pull/171#issuecomment-2315129017).

## Review guidelines

## Checklist before merging:


jjenscodee (approved):
thank you!

PR #174, opened 2024-09-11
feat(schema): minor schema revisions for draft/2-0-0

eeberhard:
## Description

After testing, this adds a field for the position of the "stop" node in the application graph. It also removes the "on_rate_violation" event trigger from the hardware, as it was decided that this is not really a transition event. Instead, rate errors will simply be part of the on_error transition for hardware in strict mode.

I regenerated the typescript declaration for the draft, which I didn't do in the last PR, so that file has more apparent changes.

## Review guidelines

## Checklist before merging:


jjenscodee (approved):
LGTM

eeberhard:
@domire8 could you confirm that you're happy with me removing the `on_rate_violation`?

domire8 (approved):
Yes that works with me sorry

PR #188, opened 2024-10-16
docs: minor improvements

domire8:
## Description

This PR updates a few lines in the docs that required a bit more precision.

## Review guidelines

## Checklist before merging:

### DIFF
--- a/docs/docs/concepts/03-ros-concepts/04-control.md
+++ b/docs/docs/concepts/03-ros-concepts/04-control.md
@@ -38,7 +38,7 @@ and outputs (GPIO).
 Command interfaces represent the desired control action that a hardware should execute. The nature of a command
 interface depends on the actuators, but commonly includes joint positions, velocities or torques. For real-time control,
 velocity and torque are the preferred command modes. For other actuators such as robot grippers, the command interface
-might instead specify a gripper distance or force through a GPIO.
+might instead specify a gripper finger distance or force through a GPIO.
 
 ## Hardware interfaces
 
--- a/docs/docs/concepts/03-ros-concepts/04-control.md
+++ b/docs/docs/concepts/03-ros-concepts/04-control.md
@@ -29,15 +29,16 @@ read the **state** of a robot and to write a **command** to the robot in a real-
 
 ### State interfaces
 
-State interfaces provide the current state of the robot. This is most commonly defined as the robot joint positions,
-and may also include joint velocities, torques or other available sensor data.
+State interfaces provide the current state of the hardware. This is most commonly defined as the robot joint positions,
+and may also include joint velocities, joint torques, other available sensor data, and general purpose digital inputs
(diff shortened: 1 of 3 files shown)


docs/docs/concepts/03-ros-concepts/04-control.md
```diff
@@ -29,15 +29,16 @@ read the **state** of a robot and to write a **command** to the robot in a real-
...
 
 ### Command interfaces
 
-Command interfaces represent the desired control action that a robot should execute. The nature of a command interface
-depends on the robot actuators, but commonly includes joint positions, velocities or torques. For real-time control,
+Command interfaces represent the desired control action that a hardware should execute. The nature of a command
+interface depends on the actuators, but commonly includes joint positions, velocities or torques. For real-time control,
 velocity and torque are the preferred command modes. For other actuators such as robot grippers, the command interface
-might instead specify a gripper distance or force.
+might instead specify a gripper distance or force through a GPIO.
```
bpapaspyros, line 41:
```suggestion
might instead specify a gripper finger distance or force through a GPIO.
```

or something more generic instead of finger

docs/docs/getting-started/01-intro.md
```diff
@@ -10,14 +10,14 @@ This guide will cover the basic steps to get started with the AICA System, from
...
-AICA Core runs in a Docker container based on Ubuntu 22.04. Thanks to containerization, the majority of the
+AICA Core runs in a Docker container based on Ubuntu 24.04. Thanks to containerization, the majority of the
 installation and usage steps in this guide will be the same for macOS or Ubuntu host platforms unless otherwise
 indicated.
 
 :::info
 
 There are some limitations with Docker on macOS compared to Docker on Ubuntu, which can affect advanced usage.
-For the best developer experience, use an Ubuntu 22.04 host machine.
+For the best developer experience, use an Ubuntu 24.04 host machine.
```
bpapaspyros, line 20:
```suggestion
For the best developer experience, use a recent Ubuntu distribution on your host machine.
```

I don't necessarily see why 24.04 would be the best. Having said that, I know that bleeding edge kernels (arch-based distros) do have some weird Docker issues. So I think it's ok to suggest Ubuntu, but not discourage the users from trying because they are for example on 20.04.

Might make more sense to mention Docker version?

domire8: Good point, do we know anything about the Docker requirements @LouisBrunner ?

LouisBrunner: Make sense for the OS. The only requirements for Docker is a somewhat recent version (one with buildx). Docker's version history is a bit tough to navigate but I imagine requiring Docker Engine v19 or later should be enough.

bpapaspyros (approved):

eeberhard (approved):

PR #190, opened 2024-10-16
feat: update timer example

domire8:
## Description

Updating the timer example to the most recent version.
This is pending a few more improvements:
- [x] Schema version 2-0-2
- [x] Core v4.0.1 that includes edge animation for the transition event
- [x] An updated GIF

## Review guidelines

## Checklist before merging:


rrosa24 (approved):

docs/docs/getting-started/04-examples/01-timer-example.md
```diff
@@ -9,49 +9,86 @@ application.
...
+        y: -20
+  buttons:
+    button:
+      display_name: Pause Timer 2
+      on_click:
+        lifecycle:
+          component: timer_2
+          transition: deactivate
+    button_2:
+      display_name: Play Timer 2
```
bpapaspyros, line 87:
```suggestion
      display_name: Start Timer 2
```

docs/docs/getting-started/04-examples/01-timer-example.md
```diff
@@ -103,50 +138,82 @@ The `events` field of a component associates component predicates with events.
...
+Once activated, the timer starts running. After 2 seconds (as specified by the `timeout` parameter), the `is_timed_out`
+predicate goes from false to true. As a result, the `transition` event causes `timer` to be unloaded and `timer_2` to be
+loaded instead. The second timer then goes through the same steps of configuring and activating before transitioning
+back to the first timer.
 
 ![timer example (animated)](./assets/timer-example.gif)
 
-Use the Pause button to keep the application in an idle state; components will remain loaded and active, but predicates
-and events will not be handled until the application is started again.
+Use the Pause and Play trigger buttons to deactivate and activate the `timer_2` while it is loaded. Verify that once
```
bpapaspyros, line 214:
```suggestion
Use the Pause and Start trigger buttons to deactivate and activate the `timer_2` while it is loaded. Verify that once
```

docs/docs/getting-started/04-examples/01-timer-example.md
```diff
@@ -103,50 +138,82 @@ The `events` field of a component associates component predicates with events.
...
+```yaml
+  buttons:
+    button:
+      display_name: Pause Timer 2
+      on_click:
+        lifecycle:
+          component: timer_2
+          transition: deactivate
+    button_2:
+      display_name: Play Timer 2
```
bpapaspyros, line 194:
```suggestion
      display_name: Start Timer 2
```

bpapaspyros (approved):
thanks! Just a few comments on whether we use `play` vs `start`

domire8:
This is ready to go now, the gif is 450% heavier than before but still only 600KB so I guess acceptable?

docs/docs/getting-started/04-examples/01-timer-example.md
```diff
@@ -103,50 +138,82 @@ The `events` field of a component associates component predicates with events.
...
-before transitioning back to the first timer.
+Once activated, the timer starts running. After 2 seconds (as specified by the `timeout` parameter), the `is_timed_out`
+predicate goes from false to true. As a result, the `transition` event causes `timer` to be unloaded and `timer_2` to be
+loaded instead. The second timer then goes through the same steps of configuring and activating before transitioning
+back to the first timer.
 
 ![timer example (animated)](./assets/timer-example.gif)
 
-Use the Pause button to keep the application in an idle state; components will remain loaded and active, but predicates
-and events will not be handled until the application is started again.
```
eeberhard, line 149: I think this is a really important point to keep, to understand what pausing the application means. It already has the potential of being confusing, so if you don't mention it but instead add a manual trigger button that pauses a timer (and not the application itself), it becomes even more confusing. I would keep this line and maybe even extend it a bit further, before introducing the idea of "pausing" (deactivating) just the timer.

eeberhard (commented):
In general I think adding the trigger buttons is doing a little bit too much at once for the first application. It forces us to talk about the specific logic of the timer (that it is paused while inactive) and at the same time it potentially creates a bit of confusion around pausing a component vs pausing the whole application. 

I would keep the button part as an addition for the second page. It can be used to introduce manipulating the graph directly to add buttons, change the display name, creating and selecting event edges. By having it as the next example it keeps the ideas nicely separate, step by step.

yrh012 (approved):
Looks good to me

jjenscodee (approved):

PR #192, opened 2024-10-18
feat(schema): add intra process comms field to component schema

domire8:
## Description

This PR adds intra process comms field to components in the application schema. I don't know if we should directly go for 2-0-3 or if it might be helpful to have a draft 2-0-3.

## Review guidelines

## Checklist before merging:

### DIFF
--- a/schemas/applications/schema/components.schema.json
+++ b/schemas/applications/schema/components.schema.json
@@ -79,6 +79,12 @@
             }
           }
         },
+        "intra_process_communications": {
+          "title": "Use Intra-process Communications",
+          "description": "If true, messages on topics which are published and subscribed to within this context will go through a special intra-process communication code code path which can avoid serialization and deserialization, unnecessary copies, and achieve lower latencies in some cases.",
+          "default": true,
+          "type": "boolean"
+        },
         "events": {
           "title": "Component Events",
           "description": "Component event sources that can trigger events",
--- a/schemas/applications/schema/components.schema.json
+++ b/schemas/applications/schema/components.schema.json
@@ -81,7 +81,7 @@
         },
         "intra_process_communications": {
           "title": "Use Intra-Process Communications",
-          "description": "If true, publishers and subscribers within the same process will enable intra-process communication which can avoid (de)serialization and unnecessary copies of messages, achieving lower latencies CPU load in some cases.",
+          "description": "If true, publishers and subscribers within the same process will enable intra-process communication which can avoid (de)serialization and unnecessary copies of messages, achieving lower latencies and CPU load in some cases",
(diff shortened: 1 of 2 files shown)


schemas/applications/schema/components.schema.json
```diff
@@ -138,6 +144,9 @@
             }
           }
         }
+      },
+      {
+
```
domire8, line 149:
```suggestion
```

domire8:
Are we interested in adding the `intra_process_communication` field to components or rather not?

schemas/applications/schema/application.schema.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$id": "/2-0-2/application.schema.json",
+  "$id": "/2-0-3-draft/application.schema.json",
```
domire8, line 2:
```suggestion
  "$id": "/2-0-2/application.schema.json",
```

domire8:
Not sure what will happen here in the CI, but it seems that there will be a PR generated automatically even though there is no version change..

schemas/applications/schema/components.schema.json
```diff
@@ -79,6 +79,12 @@
             }
           }
         },
+        "intra_process_communications": {
+          "title": "Use Intra-process Communications",
+          "description": "If true, messages on topics which are published and subscribed to within this context will go through a special intra-process communication code code path which can avoid serialization and deserialization, unnecessary copies, and achieve lower latencies in some cases.",
```
eeberhard, line 84:
```suggestion
          "title": "Use Intra-Process Communications",
          "description": "If true, publishers and subscribers within the same process will enable intra-process communication which can avoid (de)serialization and unnecessary copies of messages, achieving lower latencies CPU load in some cases.",
```

schemas/applications/schema/components.schema.json
```diff
@@ -79,6 +79,12 @@
             }
           }
         },
+        "intra_process_communications": {
+          "title": "Use Intra-Process Communications",
+          "description": "If true, publishers and subscribers within the same process will enable intra-process communication which can avoid (de)serialization and unnecessary copies of messages, achieving lower latencies CPU load in some cases.",
```
eeberhard, line 84: oops
```suggestion
          "description": "If true, publishers and subscribers within the same process will enable intra-process communication which can avoid (de)serialization and unnecessary copies of messages, achieving lower latencies and CPU load in some cases.",
```

eeberhard (approved):
Great to have it there as default true in case anyone wants to disable it for some reason, and since it's not urgent we can wait to change the version with the next change. Thanks!

PR #193, opened 2024-11-04
feat: update python api client

domire8:
## Description

This PR updates the python API client to the most recently available api endpoints. I'm not 100% sure if I miss some decorators for `@_requires_api_version('>=X.Y.Z')` or `@deprecated` or if we can even fully remove the deprecated endpoint, but appreciate if you give this a thought yourself. Thanks

## Review guidelines

## Checklist before merging:

### DIFF
--- a/python/src/aica_api/client.py
+++ b/python/src/aica_api/client.py
@@ -122,7 +122,7 @@ class AICA:
             self._logger.error(f'Error connecting to the API server at {self._address}! '
                                f'Check that AICA Core is running and configured with the right address.')
 
-        if not semver.Version.is_valid(core_version):
+        if not semver.Version.is_valid(f'{core_version}'):
             self._logger.warning(f'Invalid format for the AICA Core version {core_version}! This could be a result'
                                  f'of an internal or pre-release build of AICA Core.')
             core_version = None
--- a/python/src/aica_api/client.py
+++ b/python/src/aica_api/client.py
@@ -114,12 +114,17 @@ class AICA:
 
         version_info = semver.parse_version_info(self._api_version)
 
-        if version_info.major == 3:
+        if version_info.major == 4:
             return True
-        elif version_info.major > 3:
+        elif version_info.major > 4:
             self._logger.error(f'The detected API version v{self._api_version} is newer than the maximum API version '
                                f'supported by this client (v{self.client_version()}). Please upgrade the Python API '
                                f'client version for newer API server versions.')
+            return False
+        elif version_info.major == 3:
+            self._logger.error(f'The detected API version v{self._api_version} is older than the minimum API version '
+                               f'supported by this client (v{self.client_version()}). Please downgrade the Python API '
+                               f'client to version v2.1.0 for API server versions v3.X.')
             return False
         elif version_info.major == 2:
             self._logger.error(f'The detected API version v{self._api_version} is older than the minimum API version '
--- a/python/src/aica_api/client.py
+++ b/python/src/aica_api/client.py
@@ -1,6 +1,6 @@
 from deprecation import deprecated
 from functools import wraps
-from logging import getLogger
+from logging import getLogger, INFO
 from typing import Union, List
 
 import os
@@ -22,12 +22,13 @@ class AICA:
     """
 
     # noinspection HttpUrlsUsage
-    def __init__(self, url: str = 'localhost', port: Union[str, int] = '8080'):
+    def __init__(self, url: str = 'localhost', port: Union[str, int] = '8080', log_level = INFO):
         """
         Construct the API client with the address of the AICA application.
 
         :param url: The IP address of the AICA application
         :param port: The API port for HTTP REST endpoints (default 8080)
+        :param log_level: The desired log level
         """
         if not isinstance(port, int):
             port = int(port)
@@ -40,6 +41,7 @@ class AICA:
             self._address = f'http://{url}:{port}/api'
 
         self._logger = getLogger(__name__)
+        self._logger.setLevel(log_level)
         self._protocol = None
         self._core_version = None
 
--- a/python/src/aica_api/client.py
+++ b/python/src/aica_api/client.py
@@ -52,13 +52,13 @@ class AICA:
         return f'{self._address}/v2/{endpoint}'
 
     @staticmethod
-    def _requires_api_version(version):
+    def _requires_core_version(version):
         """
-        Decorator to mark a function with a specific API server version constraint.
+        Decorator to mark a function with a specific AICA Core version constraint.
         Elides the function call and returns None with a warning if the version constraint is violated.
 
         Example usage:
-        @_requires_api_version('>=3.2.1')
+        @_requires_core_version('>=3.2.1')
         def my_new_endpoint()
           ...
 
@@ -414,7 +414,7 @@ class AICA:
         return read_until(lambda data: data[component]['state'] == state, url=self._address, namespace='/v2/components',
                           event='component_data', timeout=timeout) is not None
 
-    @_requires_api_version('>=3.1.0')
+    @_requires_core_version('>=3.1.0')
     def wait_for_hardware(self, hardware: str, state: str, timeout: Union[None, int, float] = None) -> bool:
         """
         Wait for a hardware interface to be in a particular state. Hardware can be in any of the following states:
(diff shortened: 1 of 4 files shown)


python/src/aica_api/client.py
```diff
@@ -143,7 +148,7 @@ def controller_descriptions(self) -> requests.Response:
         """
         return requests.get(self._endpoint('controllers'))
 
-    def call_service(self, component: str, service: str, payload: str) -> requests.Response:
+    def call_component_service(self, component: str, service: str, payload: str) -> requests.Response:
```
domire8, line 151: Do I need to deprecate this endpoint the same way as it is with `wait_for_predicate`?

python/src/aica_api/client.py
```diff
@@ -114,13 +114,18 @@ def check(self) -> bool:
 
         version_info = semver.parse_version_info(self._api_version)
 
-        if version_info.major == 3:
+        if version_info.major == 4:
             return True
-        elif version_info.major > 3:
+        elif version_info.major > 4:
             self._logger.error(f'The detected API version v{self._api_version} is newer than the maximum API version '
```
eeberhard, line 120: The `/version` endpoint now returns the CORE version, not the API version (which is determined simply by the `/protocol` endpoint as v2. So the compatibility checks and tables still make sense moving forward, but we should change the language of these messages:

"The detected core version ..."

Also, we should refactor `api_version()` to be called `core_version()` instead (and in that sense mark `api_version()` as deprecated), and finally add a `self.protocol()` method to return the result of `/protocol`.

python/src/aica_api/client.py
eeberhard: It would be better for user experience yes

eeberhard (commented):
I think it would make sense to release a major update to this client after the v4 core release given the changes we made to revised /version and /protocol endpoints.

We could potentially remove the deprecated functions that are marked iwth `removed_in='3.0.0'` as part of the major update, but ones that haven't yet had a deprecated warning should get a new one before being removed entirely, just to try to keep the best user experience.

Some info and version compatibility table in the README also needs updating

python/src/aica_api/client.py
```diff
@@ -58,7 +58,7 @@ def _requires_api_version(version):
         Elides the function call and returns None with a warning if the version constraint is violated.
 
         Example usage:
-        @_requires_api_version('>=3.2.1')
+        @_requires_core_version('>=3.2.1')
```
domire8, line 61:
```suggestion
        @_requires_api_version('>=3.2.1')
```

python/src/aica_api/client.py
domire8: Should we also rename the decorator to `_requires_core_version` ?

domire8:
> Some info and version compatibility table in the README also needs updating

Will do, I needed to make sure that I got all the code changes correctly first

eeberhard (commented):
From commit e3eaa7a6ca953d0bc01f3203936ad006d9139bd4 I see you had considered renaming the static protected method `_requires_api_version` to `_requires_core_version` - I think you could safely do that as long as you fully update all references

domire8:
Now I think we are good to go

domire8:
Ok looks good to me :+1:

python/src/aica_api/client.py
```diff
@@ -37,10 +37,14 @@ def __init__(self, url: str = 'localhost', port: Union[str, int] = '5000'):
         elif '//' in url or ':' in url:
             raise ValueError(f'Invalid URL format {url}')
         else:
-            self._address = f'http://{url}:{port}'
+            if port == 5000:
```
LouisBrunner, line 40: We should make the default port `8080` no?

domire8: like this ? 507e0a7fcb51c45466b9c7c352c6cdf9724cba56

domire8:
So currently,
```python3
aica.api_version()
# >>> Error getting version details! Expected a map of `signed_packages` to include `aica_api_server`: 'aica_api_server'

aica.core_version()
# >>> {'message': 'b\'{"online":true,"signed_packages":{},"entitlements":["aica_studio_access","is_aica_developer"]}\'. You have requested this URI [/api/version] but did you mean /api//version or /api/swagger.json or /api//v2/features ?'}
```

Are those results expected and desired? Especially the second one seems weird to me

python/src/aica_api/client.py
LouisBrunner: Yes, perfect.

LouisBrunner:
I am unsure about the URL "parsing" that is happening with `startswith("http://")` and such. What is this about?

eeberhard (approved):
For me this is good to go now. Sorry for the long delay and many additional changes but it's quite a good improvement!

domire8:
Still a few problems, are we happy with that?
image

eeberhard:
> Still a few problems, are we happy with that?

🤦 looks like I made a commit error and didn't add all my changes (since I had already staged identical changes to507e0a7fcb51c45466b9c7c352c6cdf9724cba56 locally and had to rebase, I lost some parts)

I had modified the core_version function to check if the version is valid semver before returning it, I'll redo that

python/src/aica_api/client.py
```diff
@@ -81,18 +91,44 @@ def wrapper(self, *args, **kwargs):
...
-            self._logger.debug(f'API server version identified as {self._api_version}')
+            core_version = requests.get(f'{self._address}/version').json()
         except requests.exceptions.RequestException:
             self._logger.error(f'Error connecting to the API server at {self._address}! '
-                               f'Check that the AICA container is running and configured with the right address.')
-            self._api_version = None
-        return self._api_version
+                               f'Check that AICA Core is running and configured with the right address.')
+
+        if not semver.Version.is_valid(core_version):
```
domire8, line 125: This still doesn't work because the `is_valid` requires a string, and in case of a custom core, the core_version is a dict

eeberhard: ugh the problem is that I lost my previous changes and couldn't remember exactly how I did it the first time. My fault for not testing. I did actually do try / except TypeError around `parse_version_info` the first time but then thought that Version.isValid is probably nicer, forgetting that the type error will still happen

eeberhard:
Ok I actually tested it this time, sorry for the many back and forths

```python
aica.check()

print(aica.core_version())
print(aica.client_version())
print(aica.protocol())
print(aica.license().json())

print(aica.manage_sequence('a', 'start'))
```

For official v4.1.0 core
```console
4.1.0
3.0.0
v2
{'online': True, 'signed_packages': {'python_component_manager': '1.1.2', 'core/hardware-collection': '4.0.1', 'core/collection': '1.1.0', 'core/backend': '1.1.1', 'core': '4.1.0', 'aica_generic_description': '1.0.0', 'aica_event_engine_interfaces': '5.0.0', 'aica_event_engine': '5.1.0', 'aica_core_interfaces': '2.0.0', 'aica_core_controllers': '5.0.2', 'aica_core_components': '1.1.0', 'aica_api_server': '4.1.1'}, 'entitlements': ['aica_studio_access', 'is_aica_developer']}
<Response [400]>
```

For a custom build of core:
```console
Invalid format for the AICA Core version {'message': 'b\'{"online":true,"signed_packages":{},"entitlements":["aica_studio_access","is_aica_developer"]}\'. You have requested this URI [/version] but did you mean /version ?'}! This could be a resultof an internal or pre-release build of AICA Core.
Invalid format for the AICA Core version {'message': 'b\'{"online":true,"signed_packages":{},"entitlements":["aica_studio_access","is_aica_developer"]}\'. You have requested this URI [/version] but did you mean /version ?'}! This could be a resultof an internal or pre-release build of AICA Core.
None
3.0.0
v2
{'online': True, 'signed_packages': {'*': None}, 'entitlements': ['aica_studio_access', 'is_aica_developer']}
Invalid format for the AICA Core version {'message': 'b\'{"online":true,"signed_packages":{},"entitlements":["aica_studio_access","is_aica_developer"]}\'. You have requested this URI [/version] but did you mean /version ?'}! This could be a resultof an internal or pre-release build of AICA Core.
The function manage_sequence requires AICA Core version >=4.0.0, but the current Core version is unknown. The function call behavior may be undefined.
<Response [400]>
```

python/src/aica_api/client.py
```diff
@@ -81,18 +91,44 @@ def wrapper(self, *args, **kwargs):
...
+            core_version = requests.get(f'{self._address}/version').json()
         except requests.exceptions.RequestException:
             self._logger.error(f'Error connecting to the API server at {self._address}! '
-                               f'Check that the AICA container is running and configured with the right address.')
-            self._api_version = None
-        return self._api_version
+                               f'Check that AICA Core is running and configured with the right address.')
+
+        if not semver.Version.is_valid(f'{core_version}'):
+            self._logger.warning(f'Invalid format for the AICA Core version {core_version}! This could be a result'
```
eeberhard, line 126:
```suggestion
            self._logger.warning(f'Invalid format for the AICA Core version {core_version}! This could be a result '
```

PR #200, opened 2024-12-20
docs: improve signals page

domire8:
## Description

This PR makes a few updates to the signals page in the docs.

## Review guidelines

## Checklist before merging:

### DIFF
--- a/docs/docs/concepts/05-building-blocks/01-signals.md
+++ b/docs/docs/concepts/05-building-blocks/01-signals.md
@@ -16,15 +16,16 @@ Often, the message topic and message type is hardcoded within the implementation
 difficult to rearrange the network topology of ROS nodes without modifying and recompiling the node implementation
 itself.
 
+<!-- TODO: we now support all message types and not just standardized -->
 Signals then are ROS 2 publishers and subscribers with dynamically assigned topics and standardized message types.
 This makes it easy to reconfigure the signal connections between different components and controllers in the application
 graph without modifying or recompiling any source code. By using standard message types, the signal compatibility
-between components and controllers is also simplified.
+between components and controllers is also simplified. When components and controllers are connected by a signal in an
+application graph, the application interpreter will try to assert that the signals have a matching type.
 
 Additionally, ROS 2 messages are data packets, not data objects. Parsing data from a message, manipulating it and
-writing it back into a message can involve a fair amount of boilerplate code.
-
-When developing an AICA component, signals are automatically converted to and from the corresponding data object.
+writing it back into a message can involve a fair amount of boilerplate code. When developing an AICA component, signals
+are usually automatically converted to and from the corresponding data object.
 
 ## Basic signal types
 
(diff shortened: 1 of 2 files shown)


docs/docs/concepts/05-building-blocks/01-signals.md
```diff
@@ -74,24 +75,26 @@ The following state variables can be exchanged as signals:
 <!-- TODO: copy and link the markdown documentation of the state representation library directly in the programming reference -->
 
 AICA state signals are built on the
-open-source [`state_representation`](https://aica-technology.github.io/control-libraries/versions/v7.1.0/md__github_workspace_source_state_representation__r_e_a_d_m_e.html)
+open-source [`state_representation`](https://aica-technology.github.io/control-libraries/versions/9.0.1/md__2github_2workspace_2source_2state__representation_2_r_e_a_d_m_e.html)
 library for Cartesian and joint state classes in C++ and Python.
 
+<!-- TODO: all other state rep types too -->
```
domire8, line 81: To be done in a followup PR

docs/docs/concepts/05-building-blocks/01-signals.md
```diff
@@ -16,15 +16,16 @@ Often, the message topic and message type is hardcoded within the implementation
...
+application graph, the application interpreter will try to assert that the signals have a matching type.
 
 Additionally, ROS 2 messages are data packets, not data objects. Parsing data from a message, manipulating it and
-writing it back into a message can involve a fair amount of boilerplate code.
-
-When developing an AICA component, signals are automatically converted to and from the corresponding data object.
+writing it back into a message can involve a fair amount of boilerplate code. When developing an AICA component, signals
+are usually automatically converted to and from the corresponding data object.
 
 ## Basic signal types
```
domire8, line 30: Also in a follow up PR, I want to provide a table that explains what these message types are mapped to, e.g. `double` -> `Float64`

bpapaspyros: One of my concerns reading this is that we say:

> Additionally, ROS 2 messages are data packets, not data objects. Parsing data from a message, manipulating it and writing it back into a message can involve a fair amount of boilerplate code. When developing an AICA component, signals are usually automatically converted to and from the corresponding data object.

but I don't find it accurate. ROS messages are structs after all. This also feels like we are deviating from ROS language a bit (which we eventually have to but we should make a clear connection first).

I would personally prefer/suggest something more along the lines of

> ROS 2 messages are lean data structures with no additional functionality other than storing data. Therefore, parsing data from a message, manipulating it and writing it back into a message can involve a fair amount of boilerplate code. Conversely, the AICA System will typically convert these message structures to and from objects that encapsulate a lot more functionality for manipulating, transforming, and in general operating on their contents.

docs/docs/concepts/05-building-blocks/01-signals.md
```diff
@@ -74,24 +75,26 @@ The following state variables can be exchanged as signals:
...
-an AICA application. Having a reduced message definition set is important to maximizing the modularity and compatibility
-of components. When components are connected by a signal in an application graph, the application interpreter will try
-to assert that the signals have a matching type.
+The set of basic and state signal types is generally enough to cover the majority of messaging needs in an AICA
+application. Having a reduced message definition set is important to maximizing the modularity and compatibility
+of components.
 
-However, any ROS 2 message can be implemented as a signal using the `custom` signal type. As long as the custom type
-between two connected components has the same name, the application will be valid.
+However, any ROS 2 message type can be implemented as a signal in an AICA application. The syntax for doing this is
```
bpapaspyros, line 91: We say publishers/subscribers are signals, but here it feels like we now identify them with messages too. I see where we are going with this, but I think it's a bit inconsistent in its use.

bpapaspyros (commented):
some concerns/comments. 

Also, we say 

> ROS 2 topics are messages ...

which is not accurate. We could lightly edit by saying 

> ROS 2 messages flow through namespaced topics that publishers and subscribers can refer to ...

or something like that. I mention it in another comment, but I think we are deviating a bit from ROS language but that might not make much sense to ROS users (it might not affect newcomers, but it would still not portray the actual structure)

bpapaspyros:
> some concerns/comments.
> 
> Also, we say
> 
> > ROS 2 topics are messages ...
> 
> which is not accurate. We could lightly edit by saying
> 
> > ROS 2 messages flow through namespaced topics that publishers and subscribers can refer to ...
> 
> or something like that. I mention it in another comment, but I think we are deviating a bit from ROS language but that might not make much sense to ROS users (it might not affect newcomers, but it would still not portray the actual structure)

ROS uses the following:

> Topics are a vital element of the ROS graph that act as a bus for nodes to exchange messages.

> Topics are one of the main ways in which data is moved between nodes and therefore between different parts of the system

PR #203, opened 2025-01-09
fix: missing checkout action

eeberhard:
## Description

I forgot something in #202 after all :roll_eyes: I just forgot to copy it over from the test fork. Missing checkout action after splitting the workflow into separate jobs.

## Review guidelines

## Checklist before merging:


domire8 (approved):
Now that's a classic CI mistake

PR #205, opened 2025-01-10
feat(schema): use versioned common interfaces schema

eeberhard:
## Description

Following on from #202 and #203, here I add the static draft/1-0-0 interfaces schema that should have been bundled if the workflow hadn't failed. Then, as the bundled schema will subsequently exist on main branch, I update all references to use the new common interfaces schema.

This shouldn't actually trigger any workflows but once everything is on main we can check that everything is correctly referenceable and then release a new version of controller and componenent descriptions accordingly.

## Review guidelines

## Checklist before merging:


eeberhard:
I didn't see #204 but that's ok. Here I also resolve those FIXME comments that anticipated this would break at some point 😁

eeberhard:
I prefer to link to a fixed version in the docs, because that way we know the text won't be out of date with what it is referencing (if things update in the background).

PR #206, opened 2025-01-10
fix(docs): center images

domire8:
## Description

Wanted to center the images, do you know if this is going to work? According to [this](https://tw-docs.com/docs/static-site-generators/docusaurus-image-center/) it should

## Review guidelines

## Checklist before merging:

### DIFF
--- a/docs/docs/getting-started/04-examples/01-timer-example.md
+++ b/docs/docs/getting-started/04-examples/01-timer-example.md
@@ -2,6 +2,9 @@
 sidebar_position: 1
 ---
 
+import autoLifecycleEventsTimer from './assets/auto-lifecycle-events-timer.png'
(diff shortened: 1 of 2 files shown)


eeberhard:
You can run the webserver locally to check how the docs will be rendered

https://github.com/aica-technology/api/tree/main/docs#local-development

eeberhard:
I think something like this will work:

```
import autoLifecycleEventsTimer from './assets/auto-lifecycle-events-timer.png';

<div class="text--center"> 
  <img src={autoLifecycleEventsTimer} alt="Auto lifecycle events timer example" /> 
</div>
```

domire8:
> You can run the webserver locally to check how the docs will be rendered

I did not know that thanks! Looks better now indeed

eeberhard (approved):

PR #207, opened 2025-01-14
feat(python): add support for the upcoming auth update

LouisBrunner:
## Description

Resolves https://github.com/aica-technology/core-backend/issues/400

Changes:
 * Added a new `api_key` argument in the constructor
 * The client should still work the same if the user isn't using a `api_key`
 * Otherwise we will automatically create and renew the auth token using the key
 * Also added URL encoding for all path arguments
 * Added a small example for testing and demonstration
 * Code was formatted by Ruff

## Review guidelines

## Checklist before merging:

- [X] Confirm that the relevant changelog(s) are up-to-date in case of any user-facing changes

### DIFF
--- a/python/examples/with_api_key.py
+++ b/python/examples/with_api_key.py
@@ -1,0 +1,11 @@
+import os
+
+from aica_api.client import AICA
+
+client = AICA(
+    api_key=os.getenv('AICA_API_KEY'),
+)
+
+assert client.check()
+print(f'Application state: {client.get_application_state().text}')
+print(client.wait_for_component('abc', 'on'))
--- a/python/examples/with_api_key.py
+++ b/python/examples/with_api_key.py
@@ -9,4 +9,4 @@ client = AICA(
 assert client.check()
 print(f'Application state: {client.get_application_state().text}')
 print(f'Application state: {client.load_component("def").text}')
-print(client.wait_for_component('abc', 'on'))
+print(client.wait_for_component('abc', 'loaded'))
--- a/python/examples/with_api_key.py
+++ b/python/examples/with_api_key.py
@@ -8,4 +8,5 @@ client = AICA(
 
 assert client.check()
 print(f'Application state: {client.get_application_state().text}')
+print(f'Application state: {client.load_component("def").text}')
 print(client.wait_for_component('abc', 'on'))
--- a/python/src/aica_api/client.py
+++ b/python/src/aica_api/client.py
@@ -182,9 +182,6 @@ class AICA:
         core_version = None
         try:
             core_version = requests.get(self.__raw_endpoint('version')).json()
-        except requests.exceptions.InternalServerError:
-            # most likely an unsigned core
-            pass
         except requests.exceptions.RequestException:
             self._logger.error(
                 f'Error connecting to the API server at {self._address}! '
--- a/python/src/aica_api/client.py
+++ b/python/src/aica_api/client.py
@@ -102,7 +102,9 @@ class AICA:
             if self.__api_key is not None:
                 self.__create_token()
                 headers = {'Authorization': f'Bearer {self.__token}'}
-            res = requests.request(method, self._endpoint(endpoint), params=params, json=json, headers=headers, timeout=5)
+            res = requests.request(
+                method, self._endpoint(endpoint), params=params, json=json, headers=headers, timeout=5
+            )
             retry -= 1
             if res.status_code == 401:
                 if self.__api_key is None:
@@ -179,6 +181,9 @@ class AICA:
         core_version = None
         try:
             core_version = requests.get(self.__raw_endpoint('version')).json()
+        except requests.exceptions.InternalServerError:
+            # most likely an unsigned core
+            pass
         except requests.exceptions.RequestException:
             self._logger.error(
                 f'Error connecting to the API server at {self._address}! '
@@ -612,7 +617,9 @@ class AICA:
         )
 
     @_requires_core_version('>=3.1.0')
-    def wait_for_component_predicate(self, component: str, predicate: str, timeout: Union[None, int, float] = None) -> bool:
+    def wait_for_component_predicate(
+        self, component: str, predicate: str, timeout: Union[None, int, float] = None
+    ) -> bool:
         """
         Wait until a component predicate is true.
 
--- a/python/src/aica_api/client.py
+++ b/python/src/aica_api/client.py
@@ -93,7 +93,6 @@ class AICA:
         endpoint: str,
         *,
         params: Optional[dict] = None,
-        auth: Optional[bool] = None,
         json: Optional[dict] = None,
     ) -> requests.Response:
         headers = None
--- a/python/src/aica_api/client.py
+++ b/python/src/aica_api/client.py
@@ -66,7 +66,7 @@ class AICA:
 
     def __create_token(self) -> None:
         """Authenticate with the API and store the result in self.__token."""
-        if self.__token is None:
+        if self.__token is not None:
             return
         rep = requests.post(self._endpoint('auth/login'), headers={'Authorization': f'Bearer {self.__api_key}'})
         rep.raise_for_status()
--- a/python/src/aica_api/client.py
+++ b/python/src/aica_api/client.py
@@ -68,9 +68,9 @@ class AICA:
         """Authenticate with the API and store the result in self.__token."""
         if self.__token is not None:
             return
-        rep = requests.post(self._endpoint('auth/login'), headers={'Authorization': f'Bearer {self.__api_key}'})
-        rep.raise_for_status()
-        self.__token = rep.json()['token']
+        res = requests.post(self._endpoint('auth/login'), headers={'Authorization': f'Bearer {self.__api_key}'})
+        res.raise_for_status()
+        self.__token = res.json()['token']
 
     def _sio_auth(self) -> Optional[str]:
         # FIXME: doesn't handle token expiration
--- a/python/src/aica_api/client.py
+++ b/python/src/aica_api/client.py
@@ -1,28 +1,31 @@
-from deprecation import deprecated
-from functools import wraps
-from logging import getLogger, INFO
-from typing import Union, List
-
+import importlib.metadata
 import os
+import urllib.parse
+from functools import wraps
+from logging import INFO, getLogger
+from typing import List, Optional, Union
+
 import requests
 import semver
 import yaml
-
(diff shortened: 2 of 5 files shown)


python/src/aica_api/client.py
```diff
@@ -54,7 +59,56 @@ def _endpoint(self, endpoint=''):
...
+        headers = None
+        retry = 2
+        while retry > 0:
+            if self.__api_key is not None:
+                self.__create_token()
+                headers = {'Authorization': f'Bearer {self.__token}'}
+            res = requests.request(method, self._endpoint(endpoint), params=params, json=json, headers=headers, timeout=5)
+            retry -= 1
+            if res.status_code == 401:
+                if self.__api_key is None:
```
domire8, line 108: is there a `not` missing or could you explain me how this function works?

python/src/aica_api/client.py
```diff
@@ -54,7 +59,56 @@ def _endpoint(self, endpoint=''):
...
             self.protocol()
-        return f'{self._address}/{self._protocol}/{endpoint}'
+        return self.__raw_endpoint(f'{self._protocol}/{endpoint}')
+
+    def __raw_endpoint(self, endpoint: str) -> str:
+        return f'{self._address}/{endpoint}'
+
+    def __create_token(self) -> None:
+        """Authenticate with the API and store the result in self.__token."""
+        if self.__token is None:
```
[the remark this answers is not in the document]
LouisBrunner: I must have tested before I added this 😬

python/src/aica_api/client.py
LouisBrunner: If you have an API key:

* Ensure your token exists and use it in auth
* Do your request
* If it's 401, delete the token
* Retry, recreating the token (as it is now `None`)

This ensures that we retry on a 401 failure, which should mean that our token has expired and need to be recreated (i.e. first step the token existed and we just reused it).

If you don't an API key:

* Do your request
* If it's 401, exit

No need to retry anything, it won't change the outcome.

python/src/aica_api/client.py
```diff
@@ -54,7 +59,56 @@ def _endpoint(self, endpoint=''):
...
+    def __raw_endpoint(self, endpoint: str) -> str:
+        return f'{self._address}/{endpoint}'
+
+    def __create_token(self) -> None:
+        """Authenticate with the API and store the result in self.__token."""
+        if self.__token is not None:
+            return
+        rep = requests.post(self._endpoint('auth/login'), headers={'Authorization': f'Bearer {self.__api_key}'})
+        rep.raise_for_status()
+        self.__token = rep.json()['token']
```
eeberhard, line 73: nit: `res` for response rather than `rep`?
```suggestion
        res = requests.post(self._endpoint('auth/login'), headers={'Authorization': f'Bearer {self.__api_key}'})
        res.raise_for_status()
        self.__token = res.json()['token']
```

python/src/aica_api/client.py
```diff
@@ -54,7 +59,56 @@ def _endpoint(self, endpoint=''):
         """
         if self._protocol is None:
             self.protocol()
-        return f'{self._address}/{self._protocol}/{endpoint}'
+        return self.__raw_endpoint(f'{self._protocol}/{endpoint}')
+
+    def __raw_endpoint(self, endpoint: str) -> str:
+        return f'{self._address}/{endpoint}'
+
+    def __create_token(self) -> None:
```
eeberhard, line 67: nit: maybe `__ensure_token` since it only requests and creates one if it is not None?

python/src/aica_api/client.py
```diff
@@ -54,7 +59,56 @@ def _endpoint(self, endpoint=''):
...
+        """
+        return urllib.parse.quote_plus(uri)
+
+    def _request(
+        self,
+        method: str,
+        endpoint: str,
+        *,
+        params: Optional[dict] = None,
+        auth: Optional[bool] = None,
```
eeberhard, line 96: this is unused

eeberhard (commented):
Some initial comments, generally seems good! I tested with API keys with different scopes it was behaving as expected.

Before an updated release of this client I will also update the README and compatibility tables to talk about the changes to authentication, but I will do that as a followup PR.

python/src/aica_api/client.py
```diff
@@ -119,14 +179,21 @@ def core_version(self) -> Union[str, None]:
         """
         core_version = None
         try:
-            core_version = requests.get(f'{self._address}/version').json()
+            core_version = requests.get(self.__raw_endpoint('version')).json()
+        except requests.exceptions.InternalServerError:
+            # most likely an unsigned core
+            pass
```
eeberhard, line 185: If the server returns 500 it does not raise an exception here, so these lines are not necessary (tested it locally, this path was never executed)

LouisBrunner: It is with a non-signed backend as seen by @domire8 and I

domire8: I have seen the error on the server side, not client

python/examples/with_api_key.py
```diff
@@ -0,0 +1,12 @@
...
+from aica_api.client import AICA
+
+client = AICA(
+    api_key=os.getenv('AICA_API_KEY'),
+)
+
+assert client.check()
+print(f'Application state: {client.get_application_state().text}')
+print(f'Application state: {client.load_component("def").text}')
+print(client.wait_for_component('abc', 'on'))
```
domire8, line 12:
```suggestion
print(client.wait_for_component('abc', 'loaded'))
```

eeberhard (commented):
Something I didn't think about properly until now is backwards compatibility. This client version should ideally work with core versions v4.2.0 and earlier if possible. It would be nice for the new `_request` wrapper to check the `self._core_version` and skip the authorization steps if applicable. Or, maybe that check is more appropriate in `__ensure_token()`.

If that is implemented, then I think this can go on main and I will do the README / compatibility chart in a separate release PR as indicated previously.

eeberhard (approved):
Tested and working with older version of core, thanks!

python/src/aica_api/client.py
```diff
@@ -54,7 +59,93 @@ def _endpoint(self, endpoint=''):
...
+    def _safe_uri(uri: str) -> str:
+        """
+        Make a string safe for use in a URI by encoding special characters.
+
+        :param uri: The URI to sanitize
+        :return: The sanitized URI
+        """
+        return urllib.parse.quote_plus(uri)
+
+    def _request(
```
domire8, line 100: All endpoints now use this function to make requests, which in turn checks that the version is >=v4.3.0 in `ensure_token` so this is a breaking change for this client and makes the other deprecation warnings kind of obsolete no? @eeberhard @LouisBrunner

domire8:
So I have a few observations:
1. Running 4.2.0 with this version of the client, construct the client with `client = AICA()`. Works as Enrico mentioned, all good
2. Running 4.2.0 with this version of the client, construct the client with `client = AICA(api_key="...")`. For every request that I make, I get warnings
```
This function requires AICA Core version >=4.3.0, but the current AICA Core version is 4.2.0. The function call may fail due to lack of authentication.
This function requires AICA Core version >=4.3.0, but the current AICA Core version is 4.2.0. The function call may fail due to lack of authentication.
Application state: unset
This function requires AICA Core version >=4.3.0, but the current AICA Core version is 4.2.0. The function call may fail due to lack of authentication.
This function requires AICA Core version >=4.3.0, but the current AICA Core version is 4.2.0. The function call may fail due to lack of authentication.
Load component: The application is not running
This function requires AICA Core version >=4.3.0, but the current AICA Core version is 4.2.0. The function call may fail due to lack of authentication.
Wait for component: False
```
I know this is technically not the way to use the client but would be neat to avoid this.
3. Running a pre-release image of the core with this version of the client, construct the client with  `client = AICA(api_key="...")`. No endpoint apart from `core_version`, `check`, `protocol`, `api_version()` because the client doesn't log in since the first thing in `ensure_token` is a version check (that fails) and then the login endpoint is never called. Here the output of these actions:
```
>>> aica.check()
Invalid format for the AICA Core version {'message': 'Core version not found'}! This could be a result of an internal or pre-release build of AICA Core.
False
>>> aica.api_version()
Invalid format for the AICA Core version {'message': 'Core version not found'}! This could be a result of an internal or pre-release build of AICA Core.
This function requires AICA Core version >=4.3.0, but the current Core version is unknown. The function call may fail due to lack of authentication.
Invalid format for the AICA Core version {'message': 'Core version not found'}! This could be a result of an internal or pre-release build of AICA Core.
This function requires AICA Core version >=4.3.0, but the current Core version is unknown. The function call may fail due to lack of authentication.
Error getting version details! Expected a map of `signed_packages` to include `aica_api_server`: 'signed_packages'
>>> aica.core_version()
Invalid format for the AICA Core version {'message': 'Core version not found'}! This could be a result of an internal or pre-release build of AICA Core.
>>> aica.protocol()
'v2'
>>> aica.license().text
Invalid format for the AICA Core version {'message': 'Core version not found'}! This could be a result of an internal or pre-release build of AICA Core.
This function requires AICA Core version >=4.3.0, but the current Core version is unknown. The function call may fail due to lack of authentication.
Invalid format for the AICA Core version {'message': 'Core version not found'}! This could be a result of an internal or pre-release build of AICA Core.
This function requires AICA Core version >=4.3.0, but the current Core version is unknown. The function call may fail due to lack of authentication.
'{"message": "Invalid Authorization header"}\n'
>>> aica.get_application_state().text
Invalid format for the AICA Core version {'message': 'Core version not found'}! This could be a result of an internal or pre-release build of AICA Core.
This function requires AICA Core version >=4.3.0, but the current Core version is unknown. The function call may fail due to lack of authentication.
Invalid format for the AICA Core version {'message': 'Core version not found'}! This could be a result of an internal or pre-release build of AICA Core.
This function requires AICA Core version >=4.3.0, but the current Core version is unknown. The function call may fail due to lack of authentication.
'{"message": "Invalid Authorization header"}\n'
```
IMO it is required that the api client also works with pre release versions for internal development.

eeberhard:
Thanks for checking! Point 2 can definitely happen to a normal user and is inconvenient. Point 3 is also valid, though this public client is primarily intended for use with official released versions and makes no guarantees for unofficial / internal / prerelease builds. Of course where convenient we should make it as general and compatible as possible. I think for both points 2 and 3 we can use the existence (or not) of the api_key kwarg and the core_version check in a truth table to handle each case. That can be another followup before releasing a new version here.

PR #213, opened 2025-01-28
release(client): v3.1.0

eeberhard:
## Description

- #212 

Here I update the documentation for the API client and update the version to be released as v3.1.0

## Review guidelines

## Checklist before merging:

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -1,18 +1,27 @@
 # Python AICA API Client
 
-The AICA API client module provides simple functions for interacting with the API of the AICA Core.
-
 ```shell
 pip install aica-api
 ```
 
-The client can be used to easily make API calls as shown below:
+The AICA API client module provides simple functions for interacting with the AICA Core through a REST API.
 
-```python3
+Refer to https://docs.aica.tech for more information about the AICA System.
+
+## Basic usage
+
+```python
 from aica_api.client import AICA
 
 aica = AICA()
 
+if aica.check():
+    print(f"Connected to AICA Core version {aica.core_version()}")
+```
+
+The client object can be used to easily make API calls to monitor or control AICA Core. For example:
+
+```python3
 aica.set_application('my_application.yaml')
 aica.start_application()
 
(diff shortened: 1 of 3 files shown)


python/README.md
```diff
@@ -40,6 +45,49 @@ else:
...
 ```
 
+Refer to the available methods of the `AICA` client class for more advanced usage.
+
+## Network configuration
+
+By default, the API server of AICA Core is available on the default address `localhost:8080`. Depending on the network
+configuration, the URL or port number of the AICA Core instance may be different.
+
+For example, when using AICA Launcher on macOS, the API is bound to a different, randomly generated port to avoid
```
domire8, line 55: There is no other way than that?

python/README.md
```diff
@@ -49,7 +97,8 @@ Use the following compatability table to determine which client version to use.
 
 | AICA Core version | API protocol version | Matching Python client version |
 |-------------------|----------------------|--------------------------------|
-| `4.x`             | `v2`                 | `>= 3.0.0`                     |
+| `>= 4.3`          | `v2`                 | `>= 3.1.0`                     |
```
domire8, line 100: It's a detail but 3.1.0 is technically compatible with 4.X no?

domire8 (commented):
Thanks for the update. So we don't do [point 3](https://github.com/aica-technology/api/pull/207#issuecomment-2606428259) first?

python/README.md
LouisBrunner: Not at the moment. However, in a production environment, you shouldn't run Core through Launcher (IMO) so I don't think it should be a big issue for final users.

domire8 (approved):
Let's go for it, we won't be able to support pre-release images with the current setup. I will try to understand what would have to be done when I find time for it

bpapaspyros (approved):

eeberhard:
> Let's go for it, we won't be able to support pre-release images with the current setup

Thanks, I didn't find time for it yet and I do agree it's still important to fix, but I would overall prioritize a functional client for the latest official release

PR #215, opened 2025-02-04
feat(schema): add joint positions to application schema

domire8:
## Description

- https://github.com/aica-technology/core-backend/issues/423

This PR adds `joint_positions` to the application schema such that applications can define static joint configurations to be consumed by components and controllers.

## Review guidelines

## Checklist before merging:

### DIFF
--- a/schemas/applications/schema/joint_positions.schema.json
+++ b/schemas/applications/schema/joint_positions.schema.json
@@ -11,7 +11,7 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
-        "names": {
+        "joint_names": {
           "title": "Joint Names",
           "description": "The names of the joints in the same order as the positions",
           "type": "array",
@@ -27,7 +27,7 @@
         },
         "positions": {
           "title": "Joint Positions",
-          "description": "The positions of the joints in the same order as the names",
+          "description": "The positions of the joints in the same order as the names (in SI units)",
           "type": "array",
           "minItems": 1,
           "items": {
@@ -36,7 +36,7 @@
         }
       },
       "required": [
-        "names",
+        "joint_names",
         "positions"
       ]
     }
--- a/schemas/applications/schema/joint_positions.schema.json
+++ b/schemas/applications/schema/joint_positions.schema.json
@@ -17,14 +17,22 @@
           "type": "array",
           "minItems": 1,
           "uniqueItems": true,
-          "items": "string"
+          "items": {
+            "type": "string",
+            "example": [
+              "joint_1",
+              "shoulder_joint"
+            ]
+          }
         },
         "positions": {
           "title": "Joint Positions",
           "description": "The positions of the joints in the same order as the names",
           "type": "array",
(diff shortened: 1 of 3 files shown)


schemas/applications/schema/joint_positions.schema.json
```diff
@@ -0,0 +1,36 @@
...
+          "title": "Joint Names",
+          "description": "The names of the joints in the same order as the positions",
+          "type": "array",
+          "minItems": 1,
+          "uniqueItems": true,
+          "items": "string"
+        },
+        "positions": {
+          "title": "Joint Positions",
+          "description": "The positions of the joints in the same order as the names",
```
domire8, line 24: I was wondering if we should add the units here ([m] for prismatic, [rad] for revolute)

schemas/applications/schema/joint_positions.schema.json
```diff
@@ -0,0 +1,36 @@
...
+          "type": "array",
+          "minItems": 1,
+          "items": "number"
+        }
+      },
+      "required": [
```
domire8, line 30: @eeberhard In the parent issue you've mentioned that you can assert equal length, I couldn't find any hints that this is actually possible. Do you know how?

schemas/applications/schema/joint_positions.schema.json
```diff
@@ -0,0 +1,36 @@
...
+  "type": "object",
+  "additionalProperties": false,
+  "patternProperties": {
+    "^[a-z]([a-z0-9_]?[a-z0-9])*$": {
+      "title": "Joint Positions",
+      "description": "The names and positions of a static joint configuration in the application",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "names": {
```
domire8, line 14: `names` enough or do we need `joint_names`?

domire8:
@bpapaspyros also while we are here, would you like to add fields for [this issue](https://github.com/aica-technology/core-backend/issues/410)?

schemas/applications/schema/joint_positions.schema.json
eeberhard: you're right, I don't think it's easily possible. You can perform validation on the number of items in an array, and you can also use dependent schemas / conditional sub-schemas to say "if this property exists, this other condition must be met also". You can therefore say something along the lines of "if joint names exists and has 4 items, then positions must also exist with exactly 4 items", but using the length of the first property to validate the second property does not work in a general case as far as I can tell.

In any case there is also no cross-validation possible for the order of the items, and ultimately that is the agreed compromise of this representation

schemas/applications/schema/joint_positions.schema.json
eeberhard: I would prefer `joint_names`

eeberhard (commented):
I would not yet update the schema version, and I don't think it needs to be released as a draft since there aren't many complex or breaking changes here to validate. We can just check out how it feels and then release it when we're ready

PR #216, opened 2025-02-04
feat(schema): add custom tf prefix and reference frame options in hardware schema

bpapaspyros:
## Description

- https://github.com/aica-technology/core-backend/issues/426

This PR solves the issue by adding 2 new options in the hardware section of the schema. This way, we can now define a `tf_prefix` and `reference_frame` that can be internally parsed as part of the hardware.

## Review guidelines

## Checklist before merging:

### DIFF
--- a/schemas/applications/schema/hardware.schema.json
+++ b/schemas/applications/schema/hardware.schema.json
@@ -27,7 +27,7 @@
           ]
         },
         "tf_prefix": {
-            "description": "The TF prefix to use for the URDF",
+            "description": "An optional TF prefix to apply to all of the links in the URDF to distinguish multiple instances of the same robot",
             "type": "string",
             "default": "",
             "examples": [
@@ -37,9 +37,8 @@
             ]
         },
         "reference_frame": {
-            "description": "The reference frame to use for the URDF",
+            "description": "An optional reference frame to relocate the hardware in the 3D scene. If left empty, the root link in the URDF will be used as the reference frame of the hardware instead.",
             "type": "string",
-            "default": "",
             "examples": [
               "world"
             ]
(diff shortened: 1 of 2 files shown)


domire8 (commented):
Looks good to me

schemas/applications/schema/hardware.schema.json
```diff
@@ -26,6 +26,24 @@
...
+            "type": "string",
+            "default": "",
+            "examples": [
+              "my_robot_",
+              "left_robot_",
+              "robot_1_"
+            ]
+        },
+        "reference_frame": {
+            "description": "The reference frame to use for the URDF",
```
eeberhard, line 40:
```suggestion
            "description": "An optional reference frame to relocate the hardware in the 3D scene. If left empty, the root link in the URDF will be used as the reference frame of the hardware instead.",
```

schemas/applications/schema/hardware.schema.json
```diff
@@ -26,6 +26,24 @@
...
+            "type": "string",
+            "default": "",
```
eeberhard, line 42: Here I would omit the default tag; not supplying a reference frame is the same as using the root link of the URDF as the reference frame, which is different than saying `reference_frame == ""`
```suggestion
```

schemas/applications/schema/hardware.schema.json
```diff
@@ -27,7 +27,7 @@
           ]
         },
         "tf_prefix": {
-            "description": "The TF prefix to use for the URDF",
+            "description": "An optional TF prefix to apply to all of the links in the URDF to distinguish multiple instances of the same robot",
             "type": "string",
             "default": "",
```
bpapaspyros, line 32: @eeberhard here I think an empty string could make sense, but perhaps we should also add some info on what happens if this stays empty.

perhaps a short:

> An optional TF prefix to apply to all of the links in the URDF to distinguish multiple instances of the same robot. If left empty, no changes will be made to the original URDF.

PR #221, opened 2025-02-10
release(schema): application 2-0-3 and interfaces 1-0-0

eeberhard:
## Description

It's time to release some of the staged changes to the application schema. Under [schema-ver](https://snowplow.io/blog/introducing-schemaver-for-semantic-versioning-of-schemas), this is an `ADDITION` (i.e. a patch), since it is guaranteed to be compatible with the prior schema version in all cases.

I also upgraded the interfaces schema from draft/1-0-0 to 1-0-0, since it was working as intended. A followup is to update and release the component and controller descriptions to use this new officially versioned common interface, but that can only be done after 1-0-0 has been bundled and pushed.

## Review guidelines

## Checklist before merging:

### DIFF
--- a/schemas/applications/CHANGELOG.md
+++ b/schemas/applications/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 Release Versions:
 
+- [2-0-3](#2-0-3)
 - [2-0-2](#2-0-2)
 - [2-0-1](#2-0-1)
 - [2-0-0](#2-0-0)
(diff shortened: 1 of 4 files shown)


bpapaspyros (approved):

eeberhard:
Actually, sorry, I will separate the interfaces 1-0-0 from the application schema 2-0-3 release. It's better to do these things atomically.

bpapaspyros (approved):

PR #235, opened 2025-03-13
fix(schema): less restrictive frame names

domire8:
## Description

Working on a demo, we realized that we are being super restrictive with frame names in the YAML. It is for example not possible to publish a frame "test" in reference frame "UR" just because we dont allow capital letters. I quickly checked with TF, it seems that they allow any character in their frame names, I guess we should just allow numbers, letters, underscore and maybe the dash as well. Would release that as 2-0-4.

## Review guidelines

## Checklist before merging:

### DIFF
--- a/schemas/applications/schema/frames.schema.json
+++ b/schemas/applications/schema/frames.schema.json
@@ -5,7 +5,7 @@
   "type": "object",
   "additionalProperties": false,
   "patternProperties": {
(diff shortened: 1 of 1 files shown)


bpapaspyros (approved):
Thanks, good for me

eeberhard (approved):
Seems good to me! As this is more permissive than before, it will be compatible with previous schemas and could indeed be a 2-0-4 without issue.

Issue #240, opened 2025-03-24
Add state_representation overview to docs


eeberhard:
Refactor existing state representation markdown (https://github.com/aica-technology/control-libraries/blob/main/source/state_representation/README.md)

- Use Tab groups for C++ and Python code snippets
- Use KaTeX to support math equations (https://docusaurus.io/docs/markdown-features/math-equations)

PR #252, opened 2025-04-11
feat(docs): add state representation reference with equations and C++ code

eeberhard:
## Description

- #240 

This PR solves the issue by importing the state representation documentation from https://github.com/aica-technology/control-libraries/tree/main/source/state_representation

I modified it using KaTeX for the math equations and tab groups for the C++ code snippets. The idea is to add Python code snippets at a future point. 

This is also a great chance to review this particular documentation, since the last eyes on it were >2 years ago.

## Review guidelines

Estimated Time of Review: 30 minutes

## Checklist before merging:

### DIFF
--- a/docs/docs/reference/state-representation.md
+++ b/docs/docs/reference/state-representation.md
@@ -1,0 +1,1753 @@
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# State Representation
+
+This library provides a set of classes to represent **states**, which are specific data points (state variables)
+associated with a name.
+
+In robotics, the most useful types of state refer to spatial properties. The joint angles of a robot arm, the 3D
+position and velocity of a flying drone, or the measurement of an accelerometer or force-torque sensor are all examples
+of instantaneous state variables. The [`CartesianState`](#cartesian-state) and [`JointState`](#joint-state) classes
+can represent such values in an internally consistent way while providing useful functions for conversions,
+transformations and other manipulations.
+
+State classes are designed to be extensible to support more abstract state variables or properties. Another type of
+state is a **parameter**, which this library defines as a container class to for generic named variables.
+
+The following sections describe the properties of the main state classes in the library.
+
+## State
+
+The `State` base class defines the following attributes common to all derived states:
+
+- name: the name associated with the state
+- type: the specific StateType of the object
+- timestamp: an internal timestamp which refers to the last modification
+
+### Name
+
+The name is used to label a specific state instance, and can be used to disambiguate multiple states or check their
+compatibility. The name can be accessed or modified with `get_name()` and `set_name()` respectively.
+
+### State Type
+
+A state can hold different data depending on its type. The available state types are defined by the
+`state_representation::StateType` enumeration. The type field is accessible using the `get_type()` method and allows
+introspection when working with state pointers. It is a readonly property, determined by the constructor implementation.
+
+<Tabs groupId="programming-language">
+<TabItem value="c++" label="C++">
+
+```cpp
+void check_state_type(const std::shared_ptr<state_representation::State>& state) {
+  if (state->get_type() == ...) {
+    // 
+  }
+}
+```
+
+</TabItem>
+</Tabs>
+
+
+### Timestamp
+
+The timestamp of a state records when it was last modified. Any non-const operations or mutations,
+including copy constructions, will reset the timestamp.
+
+`get_age()` returns the time since last modification in seconds, while `is_deprecated(time_delay)` can be used to
+check the age of a state against a maximum time delay.
+
+### Emptiness
+
+In addition, all states have a concept of "emptiness". The `State` base class does not refer to any specific state
+variable, and therefore holds no data. Derived classes that implement state variables do hold data. If a state is empty,
+it indicates that the data is uninitialized or otherwise invalid. The `is_empty()` method can be used to check
+emptiness, or otherwise the state object can be evaluated directly.
+
+<Tabs groupId="programming-language">
+<TabItem value="c++" label="C++">
+
+```cpp
+state_representation::State state;
+
+if (state) {
+  // the state is not empty
+} else {
+  // the state is empty
+}
+```
+
+</TabItem>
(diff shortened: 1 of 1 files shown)


bpapaspyros:
on a first read that still seems very complete to me. I'll have to do a second slower pass though. We are missing the trajectory classes which I think can be here, but we also haven't used them throughout the stack yet (perhaps a future addition on a separate PR)

eeberhard:
Rather than keeping this PR as draft, I have marked the actual documentation page as unlisted so it doesn't show up yet unless you know the link to access. That way we can do followups incrementally without need to keep a staging branch.

What I would mainly like to assess in this PR is about clarity and presentation, e.g. if there is any suggested restructuring to match the site format (heading levels or splitting classes into separate pages) 

Some followups that should be separate PRs:
- Python code blocks
- Trajectory class

docs/docs/reference/state-representation.md
```diff
@@ -0,0 +1,1757 @@
...
+pose *= state; // equivalent to pose = pose * state
+```
+
+</TabItem>
+</Tabs>
+
+The time derivative of a `CartesianPose` is a `CartesianTwist`. Because `CartesianPose` represents a displacement in
+position and orientation, it can be converted into linear and angular velocity through division by a time period.
+
+Operations with time use `std::chrono::duration` types, such as `std::chrono::milliseconds`, `std::chrono::seconds`, or
```
domire8, line 584: Here we should now mention the overloaded `.integrate()` and `.differentiate()` that are much easier to use

docs/docs/reference/state-representation.md
```diff
@@ -0,0 +1,1757 @@
...
+dynamic properties:
+
+- `positions`
+- `velocities`
+- `accelerations`
+- `torques`
+
+Each state variable is represented as an N-dimensional vector (`Eigen::VectorXd`), where N is the number of joints.
+
+By design, a `JointState` most appropriately describes a serial linkage of revolute joints as found in typical
```
domire8, line 975: Are we really limited to revolute joints already here? I'm not sure

docs/docs/reference/state-representation.md
```diff
@@ -0,0 +1,1757 @@
...
+robot arms or manipulators. The values are assumed to be in standard SI units (radians, seconds, Newton-meters).
+
+### Joint names
+
+Each joint in a `JointState` collection has a name. This can make it easier to reference the state variable value of a
+specific joint in the collection.
+
+The names can be set on construction or using the `set_names()` method. As an example, a three-link robot might be
```
domire8, line 983: Maybe add reference to `get_names()`

docs/docs/reference/state-representation.md
```diff
@@ -0,0 +1,1757 @@
...
+- `twist` (`linear_velocity` and `angular_velocity`)
+- `accelerations` (`linear_acceleration` and `angular_acceleration`)
+- `wrench` (`force` and `torque`)
+
+The linear terms always come first, followed by the angular terms.
+
+Each _paired_ state variable is represented as a 6D vector (`Eigen::VectorXd(6)`), except for the pose, which is
+represented as a 7D vector (3 for `position` and 4 for `orientation`).
+
+### Reference frames
```
domire8, line 120: Maybe mention getter and setter of reference frame

docs/docs/reference/state-representation.md
```diff
@@ -0,0 +1,1757 @@
...
+  // the state is empty
+}
+```
+
+</TabItem>
+</Tabs>
+
+A state can be marked as "empty" by calling `reset()`. The state remains empty until any data is set.
+
+## Cartesian state
```
domire8, line 91: Do we somehow have to make it clear that all these derive from State so everything that we say above (name, timestamp, etc) holds here too?

domire8, line 91: We're missing clamping and scaling

domire8 (commented):
Looks good! just a few ideas, not blocking

docs/docs/reference/state-representation.md
```diff
@@ -0,0 +1,1757 @@
...
+js.set_acceleration(0.5, 1); // set the acceleration of the knee joint to 5 rad/s^2
+js.set_torque(2.0, "knee"); // set the torque of the knee joint to 2 Nm
+```
+
+</TabItem>
+</Tabs>
+
+### Joint state addition, subtraction and scaling
+
+Two `JointState` objects can be combined with addition or subtraction, provided they have the same name and joint names.
```
domire8, line 1090: This is actually not correct, they don't need the same name:
https://github.com/aica-technology/control-libraries/blob/1f6c10fd4b4b5d3e9e35ce67a10a3dc8931ba865/source/state_representation/src/space/joint/JointState.cpp#L437
however, the error message is wrong: https://github.com/aica-technology/control-libraries/blob/1f6c10fd4b4b5d3e9e35ce67a10a3dc8931ba865/source/state_representation/src/space/joint/JointState.cpp#L519

docs/docs/reference/state-representation.md
```diff
@@ -0,0 +1,1757 @@
...
+state.set_acceleration(Eigen::VectorXd::Random(6)); // update the linear and angular acceleration
+
+Eigen::VectorXd wrench = state.get_wrench(); // 6D vector of force and torque
+state.set_wrench(Eigen::VectorXd::Random(6)); // update the force and torque
+```
+
+</TabItem>
+</Tabs>
+
+### Cartesian addition and subtraction
```
domire8, line 245: We're missing scaling for Cartesian

bpapaspyros (commented):
a few more comments on text. Some are probably just personal preferences that don't need to be applied (no strong opinion there)

I guess this shouldn't include trajectory classes just yet (perhaps a separate PR for that as it may generate a bigger discussion)

I haven't checked how this looks yet, I will perhaps do a visual check soon

eeberhard:
Thanks for the useful feedback. I'll try to apply the direct suggestions and fix the most available things, and then leave bigger new sections for a followup.

bpapaspyros (approved):

PR #255, opened 2025-04-16
ci: fix docs build and deploy

domire8:
## Description

This PR solves failures in the CI for build and deploy the docs. The workflow is tested and works, proven by the page https://docs.aica.tech/docs/getting-started/installation-and-launch which contains the latest changes to the docs (docker desktop).

## Review guidelines

## Checklist before merging:


.github/workflows/build-docs.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - uses: aica-technology/.github/.github/actions/ghcr-ensure-prefix@v0.6.1
         id: ensure-image
         with:
-          image_name: aica-technology/backend-api-docs
+          image_name: aica-technology/api-server-docs
```
domire8, line 53: Just made a change here so I can comment but what is going on here? This backend-api-docs is deprecated and the api version file in this repo contains 3.0.0 which is behind too no?

eeberhard: yes you are right that these are both outdated. `api-server-docs` is the newer image and v4.3.2 is the latest. For now l would say to use `api-server-docs:latest` as the image of choice, but that we add a new issue to take care of this properly:

I think we should:
- always pull `latest` for the default API reference link
- also pull multiple older tags of the image and put them in respectively tags subdirectories so that we can include version-specific links in a dropdown
  - I think we can just manually set the latest of each prior major release, so e.g. 3.4.1, then 4.X before v5, etc, while latest is always latest
- also bundle our openAPI schema in the image so that we distribute it more easily

eeberhard (commented):
Happy for you to change the API doc version link as part of this deployment fix

domire8:
Had to give read access to the api-server-docs package to the API repo but now it's good, already built it

eeberhard (approved):

Issue #253, opened 2025-04-14
Implement the framework for the revised docs structure


eeberhard:
.

PR #257, opened 2025-04-17
refactor: docs page structure

eeberhard:
## Description

- #253 

This PR solves the issue by reordering and creating new sidebar categories and migrating some existing content 

* Put Getting Started before Concepts
* Add Examples sidebar
* Move existing examples to new sidebar and section
* Add draft and TODO for FAQ sidebar
* Move installation guide into subsection
* Move intro page and revise some transitions

Note that this goes into a staging branch (`refactor/docs`) to prevent deploying partial changes. It should allow us to work on revising the flow of pages and deciding on any needed versioning of the docs instances before deploying (i.e. whether to freeze a version of the current docs, or have multi-versioning for different sections).

Things like new examples and guides can be written to target the refactor/docs staging branch, but if immediately useful and necessary they can instead go straight on main - they should be easy enough to migrate if they are written as standalone pages.

Still, I don't want to keep a staging branch open too long. I just feel the followup revision of content structure will be easier to review once the bulk renaming / moving of these core files and directories has been done in this PR.

## Review guidelines

## Checklist before merging:


domire8:
> Is there a way to see these changes more visually?


We can try to set it up next week. Cant say how it works on Windows without doing it myself

domire8 (approved):
> whether to freeze a version of the current docs

Dont feel like this is necessary

docs/docusaurus.config.js
```diff
@@ -82,16 +88,23 @@ const config = {
                     },
                     {
                         type: "docSidebar",
-                        sidebarId: "gettingStartedSidebar",
+                        sidebarId: "examplesSidebar",
                         position: "left",
-                        label: "Getting started",
+                        label: "Examples",
```
LouisBrunner, line 93: Shouldn't this be

```suggestion
                        label: "Guides",
```

?

LouisBrunner (approved):
The new structure is great!

docs/docusaurus.config.js
eeberhard: I think it's actually better that "guides" are a subsection of "examples" in general. There will be some examples that are not fully written out as walkthroughs with step-by-step instructions. If the navbar section header is Guides, then the less verbose examples will seem out of place.

Issue #264, opened 2025-04-26
Allow null values for parameters to reset them to empty


eeberhard:
Extend the [application schema](https://github.com/aica-technology/core-backend/issues/schemas/applications/schema/common/parameters.schema.json) to include "null" in the list of supported types.

This would enable "unsetting" a parameter, which is very distinct from setting an empty string `""` or zero/falsey value `0`/`false`.

domire8:
In which situations would it be necessary to unset parameters? Is that for the frontend to work better or how would we use that in the backend?

eeberhard:
See parent issue

PR #265, opened 2025-04-26
feat(schemas): allow null values for parameters

eeberhard:
## Description

- #264 

This PR solves the issue by extending the allowed types of the parameter value property in the application schema.

## Review guidelines

## Checklist before merging:


domire8:
Is that something that you need soon, if yes we can do the frontend and core-backend update to 2-0-5 instead of 2-0-4

eeberhard:
> Is that something that you need soon,

Yes, for me it's good to go, with the release in PR #266 (to keep functional changes separate from administrative ones)

domire8 (approved):

PR #277, opened 2025-05-12
docs: add signal interoperability examples

domire8:
## Description

This PR adds examples for the signal translation components. High level overview of how it looks:
image
image

## Review guidelines

## Checklist before merging:


docs/docs/examples/core-components/signal-interoperability.md
```diff
@@ -0,0 +1,280 @@
...
+  <img src={signalToRos} alt="Signal interoperability example 1" />
+</div>
+
+With the application loaded and playing, the two components will publish the converted message on their output signal
+each time a new message is received from the hardware interface. A jsonified version of those messages can be observed
+in the live topic view.
+
+![signal ros pose](./assets/signal-ros-pose.png)
+
+![signal ros joint](./assets/signal-ros-joint.png)
```
eeberhard, line 43: These are super small images to fit all the content. I think it would be better if they were "normal" size with the majority of the message being cut-off, and then underneath an inline code block to paste the json output, so that it is automatically formatted with a horizontal scrollbar for the long text.
```console
{"header":{"stamp":{................................................................................................................................................................................................................................................
```

domire8: image

Issue #196, opened 2024-11-06
Updated application examples for core collection


eeberhard:
Basic application logic:
  - Updated timer example (with pausing and reactivating)
  - Signal processing (cartesian integration perhaps, with 2D live view)
  - Something with a condition block 

Hardware examples:
- Robot example with point attractor (still works, just needs to be checked)
- Simple sequenced motion
- Advanced motion with logic and colliders and logic

- ROS components 
  - TF to Signal and vice versa
  - interoperating with standard messages

eeberhard:
Some additional points:

- Rough drafts of documentation pages can be workshopped in google docs (i.e. under the technology folder)
- With pugin-content-pages, we can use the `draft: true` frontmatter to have pages only show up in development, not production
- With multi-instance docs, we can maintain and version categories of the docs separately 

> I don't think we should dial down on the signal types, I actually already had some questions about them

Yes I was considering to have more details on signals and signal types under examples, or programming reference, and for that reason keep the core concepts page a bit lighter

domire8:
# Status update June 19th

## Getting Started

- Introduction (what is currently https://docs.aica.tech/docs/concepts/intro) ✅ 
- Installation (what is currently the first part of Getting Started) ✅ 
  - Compatibility ✔ --> misses notes about docker on Mac, add Windows WSL
  - Licensing (terminology needs updating for System License v. Deployment Key) ✅ 
  - Installation & launch ✅ 
- AICA Studio (brief overview of main pages and features) ✅ 
  - Managing applications (open, edit, save) ✅ 
  - Application views (graph, 3d, state table, ros topics) ✅ 
  - Controlling and monitoring an application (what do Play / Pause / Stop do) ✅ 

Kudos to @eeberhard for getting most of this done, @rrosa24 too with licensing (further updates in the future likely necessary), last touchups required by @bpapaspyros and @domire8 regarding mac and windows.

---

## Core concepts

- [REVIEW] ROS 2 Concepts (migrate from original Concepts, just review since it now comes after general introduction)
  - Built on ROS 2
  - Nodes
  - ...
  - [NEW] TF (overview of transforms are currently missing but are a key concept)
- [REVISE] Signals and Events (take from Application building blocks, use more screenshots of AICA Studio, explained in context, showing real edges on a Studio application screenshot. Much less detail about signal types)

[...]

PR #312, opened 2025-06-24
feat(docs) signals and tf

domire8:
## Description

- #196 

This PR adds concept and examples for tf to signal components

## Review guidelines

## Checklist before merging:


domire8:
Sounds good, will have a look

domire8:
Before I go further with the examples, I'd be happy to get your feedback here. I like the page in isolation, but I'm having doubts if it fits here like this. It almost feels like something that should go into guides and not into building blocks. And if we do remove the 3 different ways to create frames in AICA Studio, then I'm not sure what do write about application frames here.

eeberhard:
My rule of thumb for concepts is stuff that is "evergreen" and serves as high-level reference. If it hinges on specific UI interactions for creating and managing frames, that is better suited as a guide.

I think what you have as the intro paragraph for "Frames" and then "Using frames as signals" is perfect for the concept page. The section "Frames in AICA Studio" could be directly moved to a guide. What you have is very good though!

While we are discussing this at a broader level, I think the "components/signal-interoperability" page makes less sense now, especially as I see the pattern continuing with #306. I think I would prefer for the signal interoperability concept to be a subsection of the Signals concept page. Then we just keep the Components concept page without it being a category.

Issue #314, opened 2025-06-27
Move signal interoperability


domire8:
While we are discussing this at a broader level, I think the "components/signal-interoperability" page makes less sense now, especially as I see the pattern continuing with #306. I think I would prefer for the signal interoperability concept to be a subsection of the Signals concept page. Then we just keep the Components concept page without it being a category.

_Originally posted by @eeberhard in https://github.com/aica-technology/api/issues/312#issuecomment-3013512291_

PR #317, opened 2025-07-01
feat(docs): consolidate signal interoperability into one page

domire8:
## Description

- #314 

This PR solves the issue by putting back the signal interoperability pages onto one page, more or less reverting #311 , but in better.

## Review guidelines

## Checklist before merging:


domire8:
> In general, all good. I am not sure about the screenshots and code snippets that contain the message information. I think they clutter the page and break the flow a bit. I guess the point is to show the contents of the message (?), personally I am not sure if it is super useful.

Hmm yeah I don't disagree with that statement. I think my goal was to also sublty point the user towards the live view such that it's being used in a real example. It is debatable if thats the right place for it

Issue #301, opened 2025-06-12
Major changes to descriptions schema


eeberhard:
There are many opportunities for a v2 descriptions schema. This epic groups those ideas so that can we can draft and finalize them efficiently.

### Unify component and controller schemas into a single "description" schema

Component and controller description schemas have a lot of overlap. This already prompted the creation of the common interfaces schema to reduce code duplication. Still, it remains cumbersome to maintain two schemas and also to use them appropriately on the client side.

There are very few differences between the schemas. If we only look at unique properties:

Components have:
- registration (though it's functionally the same as "plugin" for controllers)
- input collections (though not used in our current stack, nothing inherently specific to components)
- lifecycle (boolean to indicate lifecycle state)

Controllers have:
- plugin (though it's functionally the same as "registration" for controllers)
- control type (one of position, velocity, acceleration, effort)

So really, we could get away with a single description schema if we:
- unify "registration" / "plugin" into a single "class" property
- add a "type" property that is either "component", "lifecycle component" or "controller"
- add an optional control_type property that is only required if type == controller

### Move from array of interfaces to map of interfaces

Instead of:

```json
  "parameters": [
    {
      "display_name": "Image Size",
      "description": "Image resolution [width, height] in pixels",
      "parameter_name": "image_size",
      "parameter_type": "double_array",
      "default_value": [
        480,
        840
      ]
    },
    {
      "display_name": "Object Distance",
      "description": "Assumed distance from the camera to the object in meters",
      "parameter_name": "object_distance",
      "parameter_type": "double",
      "default_value": 0.6,
      "dynamic": true
    }
  ]
```

We could have:

```json
  "parameters": {
    "image_size": {
      "display_name": "Image Size",
      "description": "Image resolution [width, height] in pixels",
      "parameter_type": "double_array",
      "default_value": [
        480,
        840
      ]
    },
    "object_distance": {
      "display_name": "Object Distance",
      "description": "Assumed distance from the camera to the object in meters",
      "parameter_type": "double",
      "default_value": 0.6,
      "dynamic": true
    }
  }
```

This would be especially compact in the YAML format:

```yaml
parameters:
  image_size:
    display_name: Image Size
    description: Image resolution [width, height] in pixels
    parameter_type: double_array
    default_value: "[ 480, 840 ]"
  object_distance:
    display_name: Object Distance
    description: Assumed distance from the camera to the object in meters
    parameter_type: double
    default_value: "0.6"
    dynamic: true
```

[...]

PR #324, opened 2025-07-17
feat(schema): revised interfaces schema

eeberhard:
## Description

- First step towards #301

This PR prepares a draft/2-0-0 version of the interfaces schema. See the CHANGELOG for a summary of changes, including the breaking changes that mark this as 2-0-0.

I tested it downstream early and often to make sure that it would not generate circular or deep references when bundled, and would also not create duplicate or conflicting types when using interface generators in Typescript or Go.

In addition to making changes in the relevant files of the interfaces schema, I also generated a bundled version for the static hosting. As with all draft schemas, this static version will be subject to future changes and removal, but having it there allows us to reference it properly in the following PRs.

It addresses most of the points in #301 but I left out parameter validation for a separate PR, just so that we could review and discuss that in more detail.

## Review guidelines

Estimated Time of Review: 20 minutes

I suggest to review by commit to see the history of changes. It is otherwise difficult to test on its own, but see the followup PR with the new descriptions schema for more context.

## Checklist before merging:

### Blocks:
- #325


domire8 (approved):
LGTM, do we want to target a staging branch?

PR #325, opened 2025-07-17
feat(schema): extensions schema for unified descriptions 

eeberhard:
## Description

- Supporting #301

This PR takes the next step towards the issue by using the draft/2-0-0 interfaces schema to create a new unified extensions schema for both components and controllers.

I also generated the bundled schema for this draft/1-0-0 version for testing.

## Review guidelines

For testing, you can use the raw URL of the generated schema, or of course download it locally, and then test it out with a suitable editor. I have included an example below.

<details>
<summary>YAML example for testing</summary>

```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/aica-technology/api/1ceb47b1c76810b461917c7a007009a97a5b6cfb/docs/static/schemas/draft/1-0-0/extension.schema.json

schema: 2-0-6
name: foo
description:
  brief: My component
type: lifecycle_component
# type: controller
# control_type: position
class: my_package::MyComponent
inherits: modulo_components::Component
assignments:
  my_assignment:
    display_name: My Assignment
    description: My assignment description
    assignment_type: int
  my_state_assignment:
    display_name: My Assignment
    description: My assignment description
    assignment_type: state
    assignment_state_type: cartesian_acceleration
inputs:
  my_input:
    display_name: My Input
    description: My input description
    signal_type: bool
  my_custom_input:
    display_name: My Input
    description: My input description
    signal_type: external
    external_signal_type: geometry_msgs::msg::Pose
parameters:
  my_parameter:
    display_name: My Parameter
    description: My parameter description
    default_value: 1
    parameter_type: int
  my_null_parameter:
    display_name: My Parameter
    description: My parameter description
    default_value: null
    parameter_type: int
    optional: true
  my_state_parameter:
    display_name: My Parameter
    description: My parameter description
    default_value: null
    parameter_type: state
    optional: false
    parameter_state_type: cartesian_acceleration
services:
  my_empty_service:
    display_name: My service
    description: My service description
  my_string_service:
    display_name: My service
    description: My service description
    service_type: string
    payload_description: A string
  my_object_service:
    display_name: My service
    description: My service description
    service_type: object
    payload_description: A string array
    payload_schema:
      $schema: https://json-schema.org/draft/2020-12/schema
      type: array
      items:
        type: string
```

</details>

## Checklist before merging:

### Blocked by:
- #324

### DIFF
--- a/docs/static/schemas/draft/1-0-0/extension.schema.json
+++ b/docs/static/schemas/draft/1-0-0/extension.schema.json
@@ -64,10 +64,8 @@
         ]
       },
       "else": {
-        "not": {
-          "required": [
-            "assignment_state_type"
-          ]
+        "properties": {
+          "assignment_state_type": false
         }
       }
     },
@@ -182,10 +180,8 @@
             ]
           },
           "else": {
-            "not": {
-              "required": [
-                "optional"
-              ]
+            "properties": {
+              "optional": false
             }
           }
         },
@@ -206,10 +202,8 @@
             ]
           },
           "else": {
-            "not": {
-              "required": [
-                "parameter_state_type"
-              ]
+            "properties": {
+              "parameter_state_type": false
             }
           }
         }
@@ -271,7 +265,6 @@
           "type": "string"
         },
         "service_type": {
-          "default": "empty",
           "enum": [
             "empty",
             "string",
@@ -306,7 +299,8 @@
       },
       "required": [
         "display_name",
-        "description"
+        "description",
+        "service_type"
       ],
       "allOf": [
         {
@@ -318,11 +312,9 @@
             }
           },
           "then": {
-            "not": {
-              "required": [
-                "payload_description",
-                "payload_schema"
-              ]
+            "properties": {
+              "payload_description": false,
+              "payload_schema": false
             }
           }
         },
@@ -332,19 +324,14 @@
               "service_type": {
                 "const": "string"
               }
-            },
-            "required": [
-              "service_type"
-            ]
+            }
           },
           "then": {
             "required": [
               "payload_description"
             ],
-            "not": {
-              "required": [
-                "payload_schema"
-              ]
+            "properties": {
+              "payload_schema": false
             }
           }
         },
@@ -354,10 +341,7 @@
               "service_type": {
                 "const": "object"
               }
-            },
-            "required": [
-              "service_type"
-            ]
+            }
           },
           "then": {
             "required": [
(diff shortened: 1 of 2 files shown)


schemas/extensions/schema/extension.schema.json
```diff
@@ -0,0 +1,197 @@
...
+      "title": "Outputs",
+      "description": "The output signals (publications) of the extension.",
+      "$ref": "#/$defs/signals"
+    },
+    "parameters": {
+      "title": "Parameters",
+      "description": "The parameters declared by the extension.",
+      "type": "object",
+      "patternProperties": {
+        "^[a-z]([a-z0-9_]?[a-z0-9])*$": {
```
domire8, line 139: I believe that this is too restrictive now. We should take it from here https://github.com/aica-technology/api/pull/260/files#diff-ea064547fa142cde57f15387faffd17eceda96b3fd8f8a55a8e536609c13d552

eeberhard: Agreed, we can indeed be more permissive for parameter names and use `^[a-zA-Z0-9]([a-zA-Z0-9_.-]?[a-zA-Z0-9])*$`. Assignments will act as parameters and so they can share the same pattern.

Do you agree that for the rest (signals, predicates and services) it can stay more restrictive? For topics we should respect what the modulo SDK does, which coerces topic names to `[a-z0-9_]` but specifically requires starting with `[a-z]`, so pretty much what we have here. For predicates it's an open question but I would prefer to start more restrictive.

eeberhard: the main thing is that those pattern properties (previously defined explicitly as e.g. `parameter_name`) are used with ROS interfaces to declare and set parameters and topic names. If the description metadata for a custom component declares that is has some parameter called `"&@(#! B!"`, it will fail at runtime because ROS will not allow that parameter name (either in the component constructor if the implementation has the same name, or else through the EventEngine that tries to set that parameter)

domire8: Yeah we can't allow anything since ROS has it's own rules about admissible topic names. However, in modulo we are much stricter and we actually automatically remove characters from a string to define parameters, e.g. a user that does `add_parameter("te@ST", ....)` will end up with a parameter called `test`.

I do agree that for everything else we can leave the pattern properties as is since names of signals, predicates and service only match modulo components and no other open source components that we make a description for (+ topic names are actually that restrictive, where there are no restrictions for parameters in ROS at all i believe)

eeberhard:
Thanks for the feedback - I revised the pattern properties for assignments and parameters and used the more streamlined syntax for disallowing properties.

domire8 (approved):

PR #326, opened 2025-07-17
feat(schema): simplified schemas for type generation

eeberhard:
## Description

- Further support of #301 

As schemas can include complex conditional validation, often automatic type generation in Typescript or Go fails. One reason for failure is circular or nested refs and defs, which is addressed by more careful schema construction and bundling as in #324. I found that our current usage of `"allOf"` is the main cause of type generation failure while also only being currently used for conditional validation to group multiple "if" statements.

I decided to create a reduced variant of the schema, calling it `*.types.schema.json`, which removes all occurrences of the `"allOf"` property. This search and deletion of keys is a bruteforce approach and could be filtered down by schema name or other white/blacklists, but for now it seems to be simple and effective.

I updated the bundleHelper.js accordingly to expect the output schema name rather than the full filename, so that it could save it twice with each extension. I also updated the CI and README with this new expected behavior.

Finally I also included the draft generated types schemas for interfaces and extensions to show the behavior and simplify downstream testing.

## Review guidelines

## Checklist before merging:

<!-- Optional: link related issues
### Blocked by:
- #325

### DIFF
--- a/docs/static/schemas/draft/1-0-0/extension.types.schema.json
+++ b/docs/static/schemas/draft/1-0-0/extension.types.schema.json
@@ -64,10 +64,8 @@
         ]
       },
       "else": {
-        "not": {
-          "required": [
-            "assignment_state_type"
-          ]
+        "properties": {
+          "assignment_state_type": false
         }
       }
     },
@@ -221,7 +219,6 @@
           "type": "string"
         },
         "service_type": {
-          "default": "empty",
           "enum": [
             "empty",
             "string",
@@ -256,7 +253,8 @@
       },
       "required": [
         "display_name",
-        "description"
+        "description",
+        "service_type"
       ]
     },
     "signal": {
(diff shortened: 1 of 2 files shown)


eeberhard:
Note: a tool for Go type generation ([go-jsonschema](https://github.com/omissis/go-jsonschema)) does not support patternProperties and therefore reverts to `interface{}` for all of the top-level interfaces of the extensions schema. However, this can be addressed quite easily as a post-generation step because the sub-schemas under the properties are themselves generated just fine. Therefore replacing for example the `Parameters map[string]interface{}` with `Parameters map[string]Parameter` would yield the correct results. The ideal case of course is that the type generation tooling does this automatically for patternProperties.

An example:

```shell
go tool go-jsonschema -p extension -o gen/extension.gen.go "${API_REPO_DIR}"/docs/static/schemas/draft/1-0-0/extension.types.schema.json

function getSedString() {
    echo "s/$1 map\[string\]interface{}/$1 map\[string\]$2/g"
}

echo "Fixing Go types..."
sed -i '' "$(getSedString "Assignments" "Assignment")" gen/extension.gen.go
sed -i '' "$(getSedString "Inputs" "Signal")" gen/extension.gen.go
sed -i '' "$(getSedString "Outputs" "Signal")" gen/extension.gen.go
sed -i '' "$(getSedString "Parameters" "Parameter")" gen/extension.gen.go
sed -i '' "$(getSedString "Predicates" "Predicate")" gen/extension.gen.go
sed -i '' "$(getSedString "Services" "Service")" gen/extension.gen.go
```

domire8 (approved):

PR #328, opened 2025-07-29
docs: add motion generation

domire8:
## Description

I don't think it's my best work but we need to strike a balance between clearly written, easy to understand, and scientifically correct. Please review with that in mind.

I added this under examples because I wanted to add the point attractor example in the same document. However, this is now already quite big and might do better under the concepts -> components tab while I would then add an actual example using the joint point attractor component and point attractor component in the examples section, referring to the concepts. What do you think?

## Review guidelines

## Checklist before merging:


bpapaspyros (commented):
nice! this reads perfectly fine to me already (text-wise) 

I do agree perhaps it needs to be split, we have a few DS components now with changes in formulation that perhaps (all of them) deserve a more dedicated mention. But would that mean the Point Attractor section here would also be part of the example (where I am imagining we have some interaction with the Studio) and not this file?

docs/docs/concepts/building-blocks/motion-generation.md
```diff
@@ -0,0 +1,96 @@
...
+the strength of the attraction being proportional to the distance to the attractor. The differential equation of this DS
+can be written as
+
+$$
+\dot x = f(x) = K(x^{\ast} - x), x \in \R^n,
+$$
+
+where $x^{\ast}$ represents the attractor, and $K$ is an additional scaling constant. The figures below show the
+response of such a DS in 2D and 3D. It can be observed that all arrows point directly at the attractor and the further
+away from the attractor they are, the longer they are.
```
SprGrf, line 86:
```suggestion
away from the attractor they are, the longer they are, signifying the increased magnitude of attraction.

PR #330, opened 2025-08-04
docs: add tip for JTC module

domire8:
## Description

This PR adds a tip in the JTC guide that links to the AICAdemy module on trajectory control.
image

## Review guidelines

## Checklist before merging:

### DIFF
--- a/docs/docs/examples/guides/application-frames.md
+++ b/docs/docs/examples/guides/application-frames.md
@@ -1,5 +1,5 @@
 ---
-sidebar_position: 5
+sidebar_position: 4
 title: Application frames
 ---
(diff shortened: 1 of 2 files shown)


bpapaspyros (approved):
good idea. We can perhaps add it in the example too, but I leave it up to you

alexialechot (approved):
🙌

domire8:
image
Added it to the example too

rrosa24:
The module number is not visible to the users, so would remove the "Module 41" part but good otherwise- thanks :)

rrosa24 (approved):

PR #333, opened 2025-08-11
feat(schema): parameter validation

eeberhard:
## Description

- https://github.com/aica-technology/community/issues/37

This PR solves the first part of the issue by defining parameter validation rules in the descriptions schema.

I used JSONSchema validation for reference about which kinds of properties make sense to validate for numeric, array and string types.

In summary, validation can be one of the following:

- a list of options (i.e. would render a dropdown select)
- value range for scalar numbers (minimum and maximum, optionally exclusive or inclusive)
- length range for arrays and strings (minimum and maximum as inclusive integer indices)
- fixed length constraint for arrays and strings
- regex pattern constraint for strings

## Review guidelines

## Checklist before merging:

### DIFF
--- a/docs/static/schemas/draft/1-0-0/extension.schema.json
+++ b/docs/static/schemas/draft/1-0-0/extension.schema.json
@@ -42,6 +42,7 @@
           "type": "string"
         },
         "assignment_type": {
+          "title": "Assignment Type",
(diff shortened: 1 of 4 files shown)


domire8 (approved):
Very nice, thanks!

PR #346, opened 2025-10-02
feat(docs): add point attractor example

domire8:
## Description

Looong overdue example page for the point attractor with IK

## Review guidelines

## Checklist before merging:

### DIFF
--- a/docs/docs/examples/core-components/point-attractor.md
+++ b/docs/docs/examples/core-components/point-attractor.md
@@ -73,11 +73,11 @@ To set up this example, follow the steps below.
    :::tip
    Learn more about the `Frame to Signal` component [here](../../concepts/building-blocks/frames.md#frame-to-signal).
    :::
-3. Add the `Signal Point Attractor` to the graph and enable auto-configure and auto-activate. Then, connect the Pose
-   output of the component from step 2 with the Attractor pose input. Also connect the Cartesian state output of the
-   Robot State Broadcaster in the hardware with the Input pose input.
-4. Add the `IK Velocity Controller` to the hardware, enable auto-load and auto-activate, and connect the Output twist of
-   the Point Attractor with the Command input of the controller.
+3. Add the `Signal Point Attractor` to the graph and enable auto-configure and auto-activate. Then, connect the `Pose`
+   output of the component from step 2 with the `Attractor pose` input. Also connect the `Cartesian state` output of the
+   `Robot State Broadcaster` in the hardware with the `Input pose` input.
+4. Add the `IK Velocity Controller` to the hardware, enable auto-load and auto-activate, and connect the `Output twist`
+   of the `Point Attractor` with the `Command` input of the controller.
 5. Finally, make sure to load all components on start by creating the necessary event edges.
 
 Start the application from AICA Studio, then switch to the 3D view. Drag the frame around and observe how the robot is


rrosa24 (approved):
Looks good to me! One thing that could elevate it would be adding more of a "why" to explain in which cases this is helpful/ making it feel more concrete

docs/docs/examples/core-components/point-attractor.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Signal Point Attractor (Cartesian space)
+
+### Interfaces
+
+The component has three inputs and one output, all of type `cartesian_state`;
+
+- Input pose: The current state of the Dynamical System, e.g. the pose that should be determined to calculate the motion
+  towards the attractor.
+- Attractor pose: The attractor of the Dynamical System, e.g. the pose that the motion should end up in.
```
bpapaspyros, line 25:
```suggestion
- Attractor pose: The attractor of the Dynamical System, e.g. the pose towards which the motion guides the robot.
```

sounds a bit better to me, without saying this is inaccurate. So up to what you prefer

docs/docs/examples/core-components/point-attractor.md
```diff
@@ -0,0 +1,217 @@
...
+   [here](../guides/application-frames.md#record-frame-in-3d-scene-view).
+2. In the same application, add the `Frame to Signal` component. Open its parameter settings, turn on auto-configure and
+   auto-activate and set the required parameter `Frame` to the name of the recorded frame from step 1.
+   :::tip
+   Learn more about the `Frame to Signal` component [here](../../concepts/building-blocks/frames.md#frame-to-signal).
+   :::
+3. Add the `Signal Point Attractor` to the graph and enable auto-configure and auto-activate. Then, connect the Pose
+   output of the component from step 2 with the Attractor pose input. Also connect the Cartesian state output of the
+   Robot State Broadcaster in the hardware with the Input pose input.
+4. Add the `IK Velocity Controller` to the hardware, enable auto-load and auto-activate, and connect the Output twist of
```
bpapaspyros, line 76:
```suggestion
4. Add the `IK Velocity Controller` to the hardware, enable auto-load and auto-activate, and connect the `Output twist` of
```

another personal preference comment. I just think it's weird to capitalize "Output" but then have "twist" after a space. Bold would work too, I just think it's not as clear this refers to a reserved keyword (same above for other signals). 

At this point we might want to establish a standard way to handle this across docs. We have some comments here and there but not sure we have internal guidelines for this (correct me if I'm wrong)

domire8: I share that concern yes. I try to use code blocks normally but if I do that here, this is what we end up with and I don't find it visually attractive. Also, I would already have to use code blocks in the interfaces above in a sense.
image

I'm not sure what the best approach is, maybe others can weigh in as well. I also struggle with this accross the whole docs

domire8: The version without blocks was harder?

bpapaspyros: for me yes. I mean any visual queue that this is something reserved would make sense to me, but it's still readable as is. I see how it might not always be very nice, it's just a little more visually verbose imo. 

Again, not strong opinion though. I am also asking for my future PRs

docs/docs/examples/core-components/point-attractor.md
```diff
@@ -0,0 +1,216 @@
...
+implementations of a Point Attractor, the `Signal Point Attractor` which acts on signals in Cartesian space and the
+`Signal Joint Point Attractor` which has the same behavior in joint space.
+
+## Signal Point Attractor (Cartesian space)
+
+### Interfaces
+
+The component has three inputs and one output, all of type `cartesian_state`;
+
+- Input pose: The current state of the DS, e.g. the pose that should be determined to calculate the motion towards
```
SprGrf, line 23: Might be a personal preference, but I find both parts of the sentence a bit confusing. I would go maybe with "state of the robot", or at least after e.g., explain that this usually corresponds to the current state of the robot. "..pose that should be determined to calculate the motion.." is a bit complex I think.

docs/docs/examples/core-components/point-attractor.md
```diff
@@ -0,0 +1,216 @@
...
+  the attractor.
+- Attractor pose: The attractor of the DS, e.g. the pose towards which the motion guides the robot.
+- Base frame: This input is reserved to advanced users and can be used in cases where the DS should be expressed in a
+  moving reference frame.
+- Output twist: The linear and angular velocity generated by the DS to drive the input towards the attractor.
+
+Usually, the goal of a Point Attractor is to move a robot to a desired point in space. The predicates of the component
+help to detect when that is the case such that other events can be triggered;
```
SprGrf, line 31:
```suggestion
help to detect when that is achieved such that other events can be triggered;

File: docs/docs/examples/core-components/point-attractor.md [hunk 5]
```diff
@@ -0,0 +1,216 @@
...
+dynamic motions in AICA Studio. As described [here](../../concepts/robotics-concepts/motion-generation.md), the motion
+of a Point Attractor DS is always directed toward a point in space, the attractor. The core components come with two
+implementations of a Point Attractor, the `Signal Point Attractor` which acts on signals in Cartesian space and the
+`Signal Joint Point Attractor` which has the same behavior in joint space.
+
+## Signal Point Attractor (Cartesian space)
+
+### Interfaces
+
+The component has three inputs and one output, all of type `cartesian_state`;
```
SprGrf, line 21: I thought the ; was intentional, but it's : in the Example section. Same in some cases below.

docs/docs/examples/core-components/point-attractor.md
```diff
@@ -0,0 +1,216 @@
...
+  [here](../../concepts/robotics-concepts/motion-generation.md#point-attractor-dynamical-system), the function of the DS
+  contains an additional scaling constant $K$, represented by the linear and angular gains parameters here. The higher
+  these values are, the higher the generated velocity. Both parameters are of type vector, so it is also possible to set
+  different gains per axis, or even disabling motion along some axis entirely.
+  :::warning
+  Setting high gains can lead to fast movements and overshoot.
+  :::
+- Maximal linear and angular velocity: Before publishing the output twist, the component clamps the linear and angular
+  velocities to the values of those two parameters. This allows to increase the gains without generating an excessive
+  response, for example.
```
SprGrf, line 57:
```suggestion
response.

File: docs/docs/examples/core-components/point-attractor.md [hunk 7]
```diff
@@ -0,0 +1,216 @@
...
+- Linear and angular precision threshold: The linear and angular distances from the attractor for the input to qualify
+  as in linear range and in angular range (see predicates above).
```
SprGrf, line 59:
```suggestion
as in linear and angular range (see predicates above).

File: docs/docs/examples/core-components/point-attractor.md [hunk 8]
```diff
@@ -0,0 +1,216 @@
...
+
+## Example
+
+To set up this example, follow these steps:
+
+1. Create a new application and record the last frame of the robot of your choice as described
```
SprGrf, line 65: Last frame here implies last one in the kinematic chain I guess. I don't know if we should make it clearer.

docs/docs/examples/core-components/point-attractor.md
```diff
@@ -0,0 +1,216 @@
...
+   :::
+3. Add the `Signal Point Attractor` to the graph and enable auto-configure and auto-activate. Then, connect the Pose
+   output of the component from step 2 with the Attractor pose input. Also connect the Cartesian state output of the
+   Robot State Broadcaster in the hardware with the Input pose input.
+4. Add the `IK Velocity Controller` to the hardware, enable auto-load and auto-activate, and connect the Output twist of
+   the Point Attractor with the Command input of the controller.
+5. Finally, make sure to load all components on start by creating the necessary event edges.
+
+Start the application from AICA Studio, then switch to the 3D view. Drag the frame around and observe how the robot is
+dynamically being attracted towards the frame.
```
SprGrf, line 80:
```suggestion
dynamically attracted towards the frame.

docs/docs/examples/core-components/point-attractor.md
domire8: I don't disagree with it but would like to stay close enough to the mathematical formulation. Simplified a little but happy if you have another suggestion

PR #342, opened 2025-09-15
docs: refactor and publish object detection example

bpapaspyros:
## Description

- #341 

This PR updates the object detection example to also mirror the latest implementation of YoloExecutor

## Review guidelines

Estimated Time of Review: 15 minutes

## Checklist before merging:

- [x] Confirm that the relevant changelog(s) are up-to-date in case of any user-facing changes

### Blocked by:
- #364
- #365 
- #368 
- [x] AICA Launcher v1.4.1
- [x] advanced perception release

PR #367, opened 2025-10-21
docs: add camera calibration walkthrough

bpapaspyros:
## Description

- #365 

Adding a camera calibration walkthrough for cases where we need to get the camera intrinsics.

## Review guidelines

## Checklist before merging:

### Blocked by:
- #364 
- #366 

### Blocks:
- #342

### DIFF
--- a/docs/docs/examples/guides/camera-calibration.md
+++ b/docs/docs/examples/guides/camera-calibration.md
@@ -5,13 +5,13 @@ title: Camera calibration
 
 import cameraCalibration from './assets/camera-calibration.gif'
 
+# Camera calibration
+
 :::info
 
 This example only works on Linux hosts as it requires graphics forwarding from Docker to your host machine.
 
 :::
-
-# Camera calibration
 
 Accurate camera calibration is essential for ensuring that visual detections and measurements correspond correctly to
 real-world geometry. In many perception pipelines, critical output is expressed in image pixel coordinates. However, if


docs/docs/examples/guides/camera-calibration.md
```diff
@@ -0,0 +1,72 @@
+---
+sidebar_position: 9
```
domire8: On main, position 9 is technically already assigned. We need better ordering anyway but for now I would say you either have to shift the robot guides backwards or you just put this at the end

docs/docs/examples/guides/camera-calibration.md
```diff
@@ -0,0 +1,72 @@
...
+---
+
+import cameraCalibration from './assets/camera-calibration.gif'
+
+# Camera calibration
+
+Accurate camera calibration is essential for ensuring that visual detections and measurements correspond correctly to
+real-world geometry. In many perception pipelines, critical output is expressed in image pixel coordinates. However, if
+your camera's lens introduces distortion or if the intrinsic parameters are unknown, these pixel coordinates can deviate
+significantly from their true positions.
```
domire8:
```suggestion
significantly from their true position.
```
I would have said position singularity but I'm not sure what is gramatically correct

docs/docs/examples/guides/camera-calibration.md
```diff
@@ -0,0 +1,72 @@
...
+
+A calibration procedure allows you to determine your camera's intrinsic (focal length, optical center) and extrinsic
+(position, orientation) parameters, as well as its distortion coefficients. Once known, these parameters can be used to
+undistort images in real time, improving the spatial accuracy of detections and any downstream estimation tasks (e.g.,
+pose reconstruction, depth reasoning, or robot alignment).
+
+## AICA helpers for calibrating cameras
+
+In a typical ROS scenario, you may have a camera whose image is published as a `sensor_msgs::msg::Image` signal, a camera officially supported by an AICA collection release, or a USB
```
domire8: This is more a question but do we anticipate that one has to calibrate Realsense and Orbbec cameras? If yes, then it's also not clear how one would apply this calibration, right?

docs/docs/examples/guides/camera-calibration.md
```diff
@@ -0,0 +1,72 @@
...
+and open a terminal at the root of `docker-images`. Then:
+
+```shell
+cd camera_calibration
+```
+
+Within that folder you will find a `build-run.sh` script that, as the name suggests, will build a Docker image and run a
+container with the camera calibration software as an entrypoint.
+
+Before running the script, make sure to generate a **checkerboard** pattern (e.g., from
```
domire8: Feel that the order is a bit off, this could maybe come before the cloning?

docs/docs/examples/guides/camera-calibration.md
```diff
@@ -0,0 +1,72 @@
...
+
+After successfully executing the script, a pop-up window displaying your image stream should appear. If the window is
+not displaying the image stream, make sure that your image streaming component is running and, for non-`CameraStreamer`
+nodes, that the correct topic name is set.
+
+Move the checkerboard in various positions and orientations until the `CALIBRATE` button is no longer grayed out (and
+most of the bars are green, indicating good sample size). Once it becomes available, press on it to start computing the
+camera matrices. After it becomes available, click on the `SAVE` button to save a recording of the process. You
+will notice a `calibration` directory has been created on your host machine under `docker-image/camera_calibration` that
+contains a compressed file. The file itself contains the images that were sampled along with a yaml file containing the
```
domire8:
```suggestion
contains a compressed file. The file itself contains the images that were sampled along with a YAML file containing the
```

docs/docs/examples/guides/camera-calibration.md
bpapaspyros: hm yes, I didn't check again after the latest PR merge, thanks. I'll put those at the end, I think (for now) it's enough to have the 3 upcoming guides back-to-back

docs/docs/examples/guides/camera-calibration.md
bpapaspyros: If one is not using the official drivers for those then they would potentially have to. Our existing components (for these 2) do output the signal for camera info already.

But in the case they do, which part isn't clear? The process doesn't change, but we don't explain how they'll get to have a `sensor_msgs::msg::Image` signal. Perhaps my use of "signal" is the culprit here. We can work with any topic of this kind coming from any sort of ROS 2 node.

docs/docs/examples/guides/camera-calibration.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+Notice that the calibration square side size is in meters. If you are using AICA's `CameraStreamer` component to produce
+the image stream, the above command should already work. If you are using your own node to stream images, you will likely
+need to specify which topic the calibrator needs to subscribe to by adding the `--calibration-topic YOUR_ROS_TOPIC`
+argument.
+
+After successfully executing the script, a pop-up window displaying your image stream should appear. If the window is
+not displaying the image stream, make sure that your image streaming component is running and, for non-`CameraStreamer`
+nodes, that the correct topic name is set.
```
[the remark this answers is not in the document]
bpapaspyros: kind of shifted things around, but not necessarily directly addressed this. Let me know

docs/docs/examples/guides/camera-calibration.md
```diff
@@ -19,17 +19,38 @@ pose reconstruction, depth reasoning, or robot alignment).
 
 ## AICA helpers for calibrating cameras
```
domire8: My notes before were a bit rough, I propose this as a new title and paragraph:
```
## Calibrate camera using AICA Studio and a checkerboard

If you are using the `CameraStreamer` with a camera with unknown calibration, we recommend to follow this guide with the
code from the corresponding [`CameraStreamer` example](./camera-streamer.md). Cameras with first-party drivers (such as
[RealSense](./realsense-component.md) and [Orbbec](./orbbec-component.md)) usually provide the intrinsic parameters and
distortion coefficients and don't need to be calibrated.

:::note

Instead of using `CameraStreamer`, advanced users can follow this process with any ROS node publishing
`sensor_msgs::msg::Image` messages from a camera.

:::
```

docs/docs/examples/guides/camera-calibration.md
```diff
@@ -0,0 +1,96 @@
...
+most of the bars are green, indicating good sample size).
+- Once `CALIBRATE` becomes available, press on it to start computing the camera matrices.
+- After it becomes available, click on the `SAVE` button to save a recording of the process.
+
+Back at your host computer's filesystem, you will notice a `calibration` directory has been created on under
+`docker-image/camera_calibration` that contains a compressed file. The file itself contains the images that were sampled
+along with a YAML file containing the camera calibration information.
+
+Finally, move the YAML file into the `data` folder of your AICA configuration such that it becomes available within AICA
+containers.
```
domire8:
```suggestion
Finally, move the YAML file into the `data` folder of your AICA configuration such that it becomes available from AICA
Studio.
```

PR #369, opened 2025-10-24
feat: versioning for core-related docs

eeberhard:
## Description

- #372 

This PR introduces a multi-site plugin to split "evergreen" docs from versioned "core" docs. Then I tagged a v4 version and set up the navbars accordingly.

TODO:
- [x] fix links (docusaurus recommends to use relative links, but this doesn't seem to work for cross-site links)
- [x] add documentation to README to explain how developers should treat versioning

## Review guidelines

## Checklist before merging:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -10,6 +10,79 @@ Contributions to the docs are welcome. Each page has an "Edit this page" link at
 corresponding source file in the GitHub repository. Make changes to the file on a branch or fork and open a Pull Request
 to the `main` branch.
 
+### Versioning
+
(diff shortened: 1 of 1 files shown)


rrosa24 (commented):
Feel this was overall a nice way to handle versioning :) Echoing some of Dominic's points/ our convo in the office: 
- Would like to default to the latest version when a user is first opening the docs
- Ideally, would like more version consistency, meaning if a user is in v4 that they stay on v4 (same with all versions)--> if this is too difficult/ impossible, we can settle for making the version tag more obvious from a design perspective

PR #389, opened 2025-11-12
docs: v5 - jtc execution

SprGrf:
## Description

New v5 version of JTC execution guide. 

## Review guidelines

## Checklist before merging:

### DIFF
--- a/docs/core/examples/core-controllers/jtc-example.md
+++ b/docs/core/examples/core-controllers/jtc-example.md
@@ -197,8 +197,8 @@ tab to start the execution of the trajectory. The main view should show the robo
 :::note
 
 Payloads cannot be modified or accessed from the Graph view on the right panel or when the application is playing. To
-modify the payload, either switch back to Graph on the main view using the mini-view switcher on the bottom left, or
-modify the payload directly on the Code tab of the right panel.
+modify the payload, stop the application and either switch back to Graph on the main view using the mini-view switcher
+or modify the payload directly on the Code tab of the right panel. In this case, the graph needs to be re-generated. 
 
 :::
 


docs/core/examples/core-controllers/jtc-example.md
```diff
@@ -165,20 +165,20 @@ below:
...
 
 The application starts by loading the `Robot State Broadcaster` and `Joint Trajectory Controller` for the generic
-six-axis robot. You may already play the application.
+six-axis robot. You may already start the application.
 
-After you press play, switch to the 3D view using the corresponding button on the top right of your AICA Studio. You
-should see your robot in its default joint configuration, along with 4 Cartesian frames named
-**start, waypoint_1, waypoint_2, and waypoint_3** that form a triangle. As you may have guessed, this application
-controls the robot such that it traverses the Cartesian frames in that order. 
+After you press start, switch to the 3D view by clicking on the minimap on the bottom left. You should see your robot in
```
eeberhard: what you call minimap is the "mini-view switcher"

bpapaspyros (approved):
Thanks, for the changes and fixes. Good for me, pending @eeberhard validation

docs/core/examples/core-controllers/jtc-example.md
domire8: Up top, it should say use AICA Core version 5.0.0 or higher

SprGrf: Thought about it. But technically this refers to the earliest version they can use to have JTC available, right? So even though this page is for v5, we still inform them about the earliest available version.

domire8: No but that doesn't make sense. It's the v5 docs so it is assumed that one uses v5. You can actually remove the whole first paragraph about launcher requirements

PR #401, opened 2025-11-17
docs: v5 - orbbec

SprGrf:
## Description

V5 updates. 

## Review guidelines

## Checklist before merging:


docs/core/examples/guides/assets/orbbec-mounted-udev-volume.png
domire8: Your launcher is outdated no? Why does it show 5.0.0 as latest? I would expect 5.0.1

SprGrf:
In purpose, as in I forgot. I might have to fix it everywhere for consistency.

PR #435, opened 2026-01-07
docs: fix footer background in light mode

domire8:
## Description

This PR adapts the styling of the footer in light mode

## Review guidelines

## Checklist before merging:

### DIFF
--- a/docs/core/studio/editor.md
+++ b/docs/core/studio/editor.md
@@ -22,7 +22,7 @@ unsaved changes. Running applications will be stopped when closing the applicati
 
 Next to the X button, the cloud storage button can be used to upload and synchronize applications to cloud storage or
 downloaded them locally. This option is disabled until a user with `admin` scope connects to cloud storage in the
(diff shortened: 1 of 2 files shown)


docs/docusaurus.config.js
```diff
@@ -155,7 +154,7 @@ const config = {
               },
               {
                 label: "More Resources",
-                href: "/docs/product/Help/resources",
+                href: "/help",
```
domire8: @alexialechot is this what you wanted to link here yes?

domire8:
On main:
image

Now fixed:
image

docs/docusaurus.config.js
alexialechot: Yes, Thank you !

domire8:
> the footer black background in light mode was intentiona

Ah sorry about that, it looked like a bug to me. I'm happy to be convinced otherwise but if I already have control over light and dark mode then the dark footer feels too aggressive as a contrast in light mode in my opinion.

PR #455, opened 2026-01-28
chore(docs): use bun instead of npm

domire8:
## Description

As title suggests. We should also take care of the CI.

## Review guidelines

## Checklist before merging:

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -13,7 +13,7 @@
 		"write-translations": "docusaurus write-translations",
 		"write-heading-ids": "docusaurus write-heading-ids",
 		"typecheck": "tsc",
-		"start:docker": "bun install && bun run start -- --host 0.0.0.0"
+		"start:docker": "bun run start -- --host 0.0.0.0"
 	},
 	"dependencies": {
 		"@docusaurus/core": "^3.7.0",


.github/workflows/build-docs.yml
```diff
@@ -12,35 +12,40 @@ jobs:
...
-      - name: Setup node
-        uses: actions/setup-node@v3
+      - name: Cache Bun dependencies
+        uses: actions/cache@v4
         with:
-          node-version: "18"
-          cache: "npm"
-          cache-dependency-path: "docs/package-lock.json"
+          path: ~/.bun/install/cache
+          key: ${{ runner.os }}-bun-${{ hashFiles('docs/bun.lock') }}
```
LouisBrunner:
```suggestion
          key: ${{ runner.os }}-bun-docs-${{ hashFiles('docs/bun.lock') }}
```

Nit: just in case we ever add another NPM package in this repo.

docs/package.json
```diff
@@ -13,12 +13,14 @@
 		"write-translations": "docusaurus write-translations",
 		"write-heading-ids": "docusaurus write-heading-ids",
 		"typecheck": "tsc",
-		"start:docker": "npm install && npm start -- --host 0.0.0.0"
+		"start:docker": "bun install && bun run start -- --host 0.0.0.0"
```
LouisBrunner: Install is already done inside the Dockerfile so maybe we can simplify?

PR #464, opened 2026-02-26
docs: FANUC guide

SprGrf:
## Description

Simple guide on how to control FANUC robots. 

## Review guidelines

## Checklist before merging:


domire8:
> I think it redirects you to the main page and that's a bit frustrating, not sure if there is a way to fix that.

There is no fallback setting we can put in place unfortunately. It is a good point though. There are already some guides that don't have v4 versions so let's merge here without the v4 version and then I open a PR to tidy v4 where we can discuss the best practices for these cases. It's out of scope for the PR at hand

domire8 (approved):

Issue #479, opened 2026-03-20
docs: add camera frame calibration guide


SprGrf:
Our `core_vision_components`, under [`dynamic_components`](https://github.com/aica-technology/dynamic-components), contain the `Robot Camera Calibration` component, which aims to solve the hand-eye calibration problem: how do I get a robot and a camera have a common reference (more info [here](https://www.mathworks.com/help/vision/ug/what-is-robot-hand-eye-calibration.html)). This package is released and can be added in Launcher configurations (under `components/core-vision`).

What we need is a guide in our Documentation page that describes how to use the component in an application context. I imagine we will need to describe something like the following:

- How does the calibration component work, what are the different options and parameters (for example, camera attached to the robot or fixed in space) 
- What markers do I need, the options I have in AICA Studio to get them (stag, aruco, or something else)
- How do I piece everything together and calibrate my camera in AICA Studio (image source, marker detection, calibration)

The above is just a suggestion, the order and actual contents can be extended further and rearranged. 

fyi: @domire8, @yrh012, @eeberhard

MithraGhlm:
Based on the discussion with @domire8 in order to have more clarity on the task, the aim is to document the procedure of calibrating the robot and camera using the aica core-vision component, so that the clients don't have to suffer :)

The steps are:
1. Gaining some knowledge about calibration and performing the intrinsic calibrating of the camera.
2. How to use Stag or Aruco marker to get the tracking component work. (this could be an stand-alone doc)
3. Calibrating the robot.

PR #482, opened 2026-04-21
docs: fiducial markers guide

MithraGhlm:
## Description

Initial draft of documentation for camera frame calibration guide.
This PR closes the camera frame calibration guide https://github.com/aica-technology/api/issues/479

This PR solves the issue by adding two stand-alone guides to the documentation. 
The guide for intrinsic camera calibration already exists.
One guide is about the configuration and the process of recognizing STag marker (WIP) and a second one would be dedicated to calibrating the robot.

## Review guidelines

## Feedback Needed
- Structure okay?
- Level of detail appropriate?


docs/core/examples/guides/camera-streamer.md
```diff
@@ -41,7 +41,7 @@ By this point, you should have something like the following:
   <img src={cameraStreamerExample} alt="Default CameraStreamer configuration" />
 </div>
 
-Before pressing Start, let us go through the parameters first. You should see:
+Before pressing Start, let's go through the parameters first. You should see:
```
SprGrf: I would keep the original, I think it's more official for documentation.

docs/core/examples/guides/stag-marker-detection.md
```diff
@@ -0,0 +1,84 @@
...
+---
+
+
+import stagDetectorExample from './assets/stag-detector-example.png'
+import stagMarkerDetection from './assets/stag-marker-detection.webm'
+
+
+# STag marker detection
+
+As a step in the Hand-Eye calibration, we need to be able to detect markers with our camera. The AICA's `core-vision` package gives you the choice between using the STag or Aruco markers.
```
SprGrf: I think it would be better to decouple this guide from Hand-Eye calibration, as this is only one example use case for the markers. I would add something more generic as an intro, and of course reference this guide on the HEC one.

docs/core/examples/guides/stag-marker-detection.md
```diff
@@ -0,0 +1,84 @@
...
+import stagDetectorExample from './assets/stag-detector-example.png'
+import stagMarkerDetection from './assets/stag-marker-detection.webm'
+
+
+# STag marker detection
+
+As a step in the Hand-Eye calibration, we need to be able to detect markers with our camera. The AICA's `core-vision` package gives you the choice between using the STag or Aruco markers. 
+
+
+:::TIP
```
SprGrf: similar below
```suggestion
:::tip

File: docs/core/examples/guides/stag-marker-detection.md [hunk 4]
```diff
@@ -0,0 +1,84 @@
...
+## Using the STag detector
+
+Launch AICA Studio with a configuration that contains the `core-vision` package and create a new application.
+
+1. Remove the hardware interface that is included in new applications by default.
+2. From the `Scene` menu, use the `Add Component` tab and look for the **Camera Streamer** and **STag Detector** components, either by searching
+or by manually going under the `Core Vision Components` menu. Add both of them to the graph.
+3. Next, connect both components to the start block. Morover, connect the outputs of the Camera Streamer to the relevant inputs of the STag Detector.
+4. Enable **auto-configure** and **auto-activate** on both components.
+5. By chosing any of the components, you can find all the available component parameters.
```
SprGrf: or something similar
```suggestion
5. By selecting any of the components, you can find all the available component parameters in the right panel under Settings.

File: docs/core/examples/guides/stag-marker-detection.md [hunk 5]
```diff
@@ -0,0 +1,84 @@
...
+
+
+<div class="text--center">
+  <img src={stagDetectorExample} alt="CameraStreamer configuration with STagDetector component" />
+</div>
+
+
+:::info
+
+The Camera Streamer parameters are explained in the [CameraStreamer component guid](./camera-streamer.md).
```
SprGrf:
```suggestion
The Camera Streamer parameters are explained in the [CameraStreamer component guide](./camera-streamer.md).

File: docs/core/examples/guides/stag-marker-detection.md [hunk 6]
```diff
@@ -0,0 +1,84 @@
...
+
+:::
+
+
+Let's go through the parameters of the STag Detector component:
+
+- **Rate**: The frequency of the callbacks.
```
SprGrf: As mentioned in the description of the component, this parameter does not affect the component. The detection happens when a new input (image) is received. You can check the phrasing for the equivalent parameter in camera streamer.

docs/core/examples/guides/stag-marker-detection.md
```diff
@@ -0,0 +1,84 @@
...
+The Camera Streamer parameters are explained in the [CameraStreamer component guid](./camera-streamer.md).
+
+:::
+
+
+Let's go through the parameters of the STag Detector component:
+
+- **Rate**: The frequency of the callbacks. 
+- **Bundle file**: 
+- **Marker selection**: The name(s) of the marker(s) that we intend to be recognized. If any of these markers enters the camera frame, the `is_any_selected_marker_detected` predicate lights up. This name should always be written with the `stage_` prefix.
```
SprGrf: Did know about the second part at all
```suggestion
- **Marker selection**: The name(s) of the marker(s) that we intend to be recognized. If any of these markers enters the camera frame, the `is_any_selected_marker_detected` predicate is set to **True**. This name should always be prepended with `stage_`.

File: docs/core/examples/guides/stag-marker-detection.md [hunk 8]
```diff
@@ -0,0 +1,84 @@
...
+- **Marker size**: 
+- **Library**: This is the ID number of the HD library utilized by STag markers. The allowed numbers are  `[11, 13, 15, 17, 19, 21, 23]`.
+- **Error correction**: 
+- **Prefix**: This prefix is used for marker names.
+
+
+:::TIP
+
+If a decision needs to be made based on the existance of a specific STag marker in the camera frame, it’s name should be indicated in the `Marker Selection` variable.
```
SprGrf:
```suggestion
If a decision needs to be made based on the existance of a specific STag marker in the camera frame, its name should be indicated in the `Marker Selection` variable.

File: docs/core/examples/guides/stag-marker-detection.md [hunk 9]
```diff
@@ -0,0 +1,84 @@
...
+
+:::
+
+
+:::info
+
+If any marker is detected in the camera frame at all, the `is_any_marker_detected` predicate will light up.
```
SprGrf: as before

SprGrf (commented):
Thanks! First quick review:

docs/core/examples/guides/stag-marker-detection.md
MithraGhlm: Yes, `Rate` doesn't have any effect on the component, but we still need to explain all the parameters I suppose, , regardless of their effects, right?

docs/core/examples/guides/stag-marker-detection.md
MithraGhlm: Actually the part of parameter explanation is left incomplete (and mostly wrong),  because I still don't know them completely.

MithraGhlm:
@SprGrf  Thanks

MithraGhlm:
Should there be a stand alone guide to introduce the `Robot Camera Calibration` component?
Or is it better to have it in the same docs page?

docs/core/examples/guides/hand-eye-calibration.md
```diff
@@ -0,0 +1,63 @@
...
+---
+// Should the name of the document change to "robot calibration" ?
+
+
+import RobotCalibrationConfiguration from './assets/robot-calibration-configuration.png'
+
+
+# Hand-Eye calibration
+
+Robot calibration is a fundamental prerequisite for any robotic system that relies on precise coordination between a vision sensor and a manipulator. It establishes the spatial transformation between the robot’s end-effector and the camera frame, enabling accurate mapping between observed features and actionable robot coordinates. Without proper calibration, even high-quality perception or motion planning algorithms can yield significant positioning errors.
```
SprGrf: Commenting here, but probably needed in the marker guide as well. We use formatters to make sure the style is consistent. Prettier might be an option, there are also more extensions in VSCode that you can use.

docs/core/examples/guides/hand-eye-calibration.md
```diff
@@ -0,0 +1,63 @@
...
+
+The AICA's `core-vision` package provides a structured workflow for performing hand–eye calibration efficiently and reproducibly.
+
+Accurate hand–eye calibration is critical in tasks such as visual servoing, object manipulation, inspection, and assembly. This tool is designed to minimize discrepancies and provide reliable calibration outputs suitable for industrial environments.
+
+
+## Robot Calibration Using AICA Studio and a Marker
```
SprGrf:
```suggestion
## Robot Calibration using AICA Studio and a marker

File: docs/core/examples/guides/hand-eye-calibration.md [hunk 12]
```diff
@@ -0,0 +1,63 @@
...
+
+After completing the camera calibration as described in the [`Camera Calibration` example](./camera-calibration.md), and verifying marker detection as outlined in the [`Marker detection`](./marker-detection.md) section, you can proceed with the hand–eye calibration process.
+
+
+This example demonstrates the eye-in-hand configuration (camera mounted on the robot arm). The procedure for the eye-to-hand configuration (static camera) follows a similar workflow.
```
SprGrf: I would add a comment about the option in the component. It would be nice if we explained all component parameters, as you did for the marker detection. The number of images that you mention below for example (100), and the path, are also configurable.

docs/core/examples/guides/hand-eye-calibration.md
```diff
@@ -0,0 +1,63 @@
...
+- Once 100 unique images have been captured, the system automatically generates a calibration file in YAML format in the following directory:
+
+```bash
+/tmp/calibration/camera_calibration.yaml
+```
+
+
+<div class="text--center">
+  <img src={RobotCalibrationConfiguration} alt="The configuration required for hand-eye calibration" />
+</div>
```
SprGrf: Starting adding a small radius lately
```suggestion
<div class="text--center">
  <img src={RobotCalibrationConfiguration} alt="The configuration required for hand-eye calibration" style={{ borderRadius: "8px" }}/>
</div>

File: docs/core/examples/guides/hand-eye-calibration.md [hunk 14]
```diff
@@ -0,0 +1,63 @@
...
+
+
+::: info
+
+The robot TCP can be moved in different ways. Using teach pendant (jogging or Freedrive mode), or using AICA Studio's Hand Guiding Controller as described in [`Hand Guiding controller`](./ur-harware-interface.md)
```
SprGrf: The hand guiding is only for UR currently. I am not sure you can use the teach pendant either. @domire8 any suggestions here?

docs/core/examples/guides/hand-eye-calibration.md
```diff
@@ -0,0 +1,63 @@
...
+## Robot Calibration Using AICA Studio and a Marker
+
+After completing the camera calibration as described in the [`Camera Calibration` example](./camera-calibration.md), and verifying marker detection as outlined in the [`Marker detection`](./marker-detection.md) section, you can proceed with the hand–eye calibration process.
+
+
+This example demonstrates the eye-in-hand configuration (camera mounted on the robot arm). The procedure for the eye-to-hand configuration (static camera) follows a similar workflow.
+
+
+- Ensure that the camera is properly configured and operational.
+- Connect all required components and controllers in AICA Launcher application. Refer to the system setup illustration below for guidance.
```
SprGrf: The application is not restricted to Launcher, we usually refer to them as AICA applications.
```suggestion
- Connect all required components and controllers in the AICA application. Refer to the system setup illustration below for guidance.

File: docs/core/examples/guides/hand-eye-calibration.md [hunk 16]
```diff
@@ -0,0 +1,63 @@
...
+- Place the marker within the robot workspace, ensuring it is fully visible to the camera. To monitor the live camera feed, enable **Launch RViz** from the Launcher settings, as described in the [`marker detection`](./marker-detection.md) guide.
+- Run the program and move the robot TCP (Tool Center Point) to capture images of the marker from multiple perspectives. The application starts capturing images automatically.
```
SprGrf: Here we need some more details. The robot has to move in way that we will be able to collect the tool transformation from tf, if I am not mistaken. So we need a controller and an app that will move the robot. This is connected to the comment below, I will leave it up to @domire8 to offer a suggestion, not sure how specific we want to be here.

SprGrf (commented):
Nice! The backbone is here, some first comments I had while reading:

docs/core/examples/guides/hand-eye-calibration.md
MithraGhlm: Yes, that's why I asked (a little bit further up this page) if we should have a stand alone guide to introduce the Robot Camera Calibration component (like CameraStreamer component that we already have).
Or you find it better if it is integrated it in this very guide?

But yes, I should definitely not mention the number of images here.

docs/core/examples/guides/hand-eye-calibration.md
MithraGhlm: Oh yes true. I should change the related image also.
I can put the hand guiding option for the UR in a separate box, as extra info.

Actually once I ran into singularity when using UR hand guide, and then I had to continue with the teach pendant. I used Jogging, and finished the sampling process. It didn't seem to have any problems with it (at least none that I'm aware of).

SprGrf: We can leave the image and still use the UR with hand guiding (we have to use a robot anyway, and it's nice to link to the rest of the guides), but it has to be clear what needs to happen. 

> I used Jogging, and finished the sampling process. It didn't seem to have any problems with it (at least none that I'm aware of).

That is because we keep getting the robot state at this point, although we are not capable to command it. Not sure this is the indicated way to do that though.

docs/core/examples/guides/hand-eye-calibration.md
SprGrf: Ah now I understand the point of the question. Yes I think this is the place to explain the component.

docs/core/examples/guides/marker-detection.md
```diff
@@ -0,0 +1,97 @@
+---
+sidebar_position: 13
```
domire8:
```suggestion
sidebar_position: 14
```

docs/core/examples/guides/hand-eye-calibration.md
```diff
@@ -0,0 +1,63 @@
+---
+sidebar_position: 14
```
domire8:
```suggestion
sidebar_position: 15
```

docs/core/examples/guides/camera-streamer.md
```diff
@@ -1,5 +1,5 @@
 ---
-sidebar_position: 13
+sidebar_position: 15
```
domire8:
```suggestion
sidebar_position: 13
```
Feels like the Camera Streamer component needs to be visited before marker detection and calibration so I would leave it in front

docs/core/examples/guides/marker-detection.md
```diff
@@ -0,0 +1,97 @@
...
+
+import stagDetectorExample from './assets/stag-detector-example.png'
+import stagMarkerDetection from './assets/stag-marker-detection.webm'
+
+
+# Marker detection
+
+Different types of markers are used in robotics to provide precise 3D pose estimation and identification for cameras, enabling or improving robotic calibration and object manipulation.
+
+The AICA's `core-vision` package gives you the choice between using two of the most famous markers, the STag marker and the Aruco markers.
```
domire8:
```suggestion
AICA's `core-vision` package gives you the choice between using two commonly used markers, the STag and Aruco. 
```

docs/core/examples/guides/marker-detection.md
```diff
@@ -0,0 +1,97 @@
...
+sidebar_position: 13
+title: Marker detection
+---
+
+
+import stagDetectorExample from './assets/stag-detector-example.png'
+import stagMarkerDetection from './assets/stag-marker-detection.webm'
+
+
+# Marker detection
```
domire8: Both Aruco and STag are *Fiducial Markers*, I would prefer to name this guide like this and also refer to them as fiducial markers

docs/core/examples/guides/marker-detection.md
```diff
@@ -0,0 +1,97 @@
...
+
+Different types of markers are used in robotics to provide precise 3D pose estimation and identification for cameras, enabling or improving robotic calibration and object manipulation.
```
domire8:
```suggestion
Different types of fiducial markers are used in robotics to provide precise 3D pose estimation and identification for cameras, enabling or improving robotic calibration and object manipulation.
```

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,97 @@
...
+# Marker detection
+
+Different types of markers are used in robotics to provide precise 3D pose estimation and identification for cameras, enabling or improving robotic calibration and object manipulation.
+
+The AICA's `core-vision` package gives you the choice between using two of the most famous markers, the STag marker and the Aruco markers. 
+
+
+:::tip
+
+It's important to notice that the marker detection should be done after the intrinsic calibration of the camera which is explained in the [calibration guide](./camera-calibration.md).
```
domire8:
```suggestion
Performing the [intrinsic calibration](./camera-calibration.md) of the camera improves the precision for fiducial marker detection and tracking.
```

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,97 @@
...
+
+:::
+
+
+The following provides an example of STag marker detection. Using the Aruco marker follows a very similar process.
```
domire8:
```suggestion
This guide provides an example of STag marker detection. Using the Aruco marker follows a very similar process.
```

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,97 @@
...
+
+
+## Using the STag detector
+
```
domire8, line 56: Feels like this is missing a section where we explain how to obtain download and print markers. Having that, it will then also be easier to explain the marker size and library parameters

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,97 @@
...
+Launch AICA Studio with a configuration that contains the `core-vision` package and create a new application.
+
+1. Remove the hardware interface that is included in new applications by default.
+2. From the `Scene` menu, use the `Add Component` tab and look for the **Camera Streamer** and **STag Detector** components, either by searching
+or by manually going under the `Core Vision Components` menu. Add both of them to the graph.
+3. Next, connect both components to the start block. Morover, connect the outputs of the Camera Streamer to the relevant inputs of the STag Detector.
```
domire8:
```suggestion
3. Next, connect both components to the start block. Moreover, connect the outputs of the Camera Streamer to the relevant inputs of the STag Detector.
```

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,97 @@
...
+4. Enable **auto-configure** and **auto-activate** on both components.
+5. By selecting any of the components, you can find all the available component parameters in the right panel under Settings.
+
```
domire8, line 70: You already had the tip above about the calibration, I think it's worth mentioning here that one should then not forget to add the filepath for it

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,97 @@
...
+:::info
+
+The Camera Streamer parameters are explained in the [CameraStreamer component guide](./camera-streamer.md).
+
+:::
+
+
+Let's go through the parameters of the STag Detector component:
+
+- **Rate**: This is the component's rate, but it has no effect on the operation of `STag Detector`.
```
domire8:
```suggestion
- **Rate**: The rate parameter doesn't affect the behavior of the component as the detection process
    occurs on reception of a new image.
```

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,97 @@
...
+- **Bundle file**: 
+- **Marker selection**: The name(s) of the marker(s) that we intend to be recognized. If any of these markers enters the camera frame, the `is_any_selected_marker_detected` predicate is set to **True**. This name should always be prepended with `stage_`.
```
domire8: It is not correct to say that the name should be prepended with `stag_` right? It should be prepended with the value of "Prefix"

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,97 @@
...
+- **Marker size**:
```
domire8: This is easy, one needs to measure the size of the printed marker

docs/core/examples/guides/hand-eye-calibration.md
domire8: Yes that is a good point, in a first draft we can provide an app with the UR Hand guiding controller.

docs/core/examples/guides/hand-eye-calibration.md
domire8: I would focus on a full example with the UR Hand guiding controller first. It doesn't apply to all robots that is clear, but it's better than nothing for now

docs/core/examples/guides/marker-detection.md
SprGrf: Something off in the suggestion

docs/core/examples/guides/fiducial-markers.md
SprGrf: This might have been my bad, did not check before making the suggestion.

docs/core/examples/guides/hand-eye-calibration.md
MithraGhlm: You mean some predefined points in the space  that the arm moves to? And then we need to have a component for it.

I'll appreciate more detail on the expected app. Some guidance to make it clear what is exactly meant by it.

SprGrf: We will provide an example app with the hand guiding controller already set up. No need for predefined points that way. 

You can check [this](https://docs.aica.tech/core/examples/guides/ur-harware-interface) example on how to do it. There we provide an example app for remote control.

docs/core/examples/guides/fiducial-markers.md
MithraGhlm: True

docs/core/examples/guides/fiducial-markers.md
MithraGhlm: @domire8  Could you pls help me with the correct definition of the `Bundle file`?

MithraGhlm: > This is easy, one needs to measure the size of the printed marker

I don't get what you mean! Which parameter?

domire8: Let's ignore bundle files for now it would be slightly out of scope. It will require its own section.

I mean parameter `Marker size`

docs/core/examples/guides/hand-eye-calibration.md
```diff
@@ -0,0 +1,63 @@
...
+The AICA's `core-vision` package provides a structured workflow for performing hand–eye calibration efficiently and reproducibly.
+
+Accurate hand–eye calibration is critical in tasks such as visual servoing, object manipulation, inspection, and assembly. This tool is designed to minimize discrepancies and provide reliable calibration outputs suitable for industrial environments.
+
+
+## Robot Calibration using AICA Studio and a marker
+
+After completing the camera calibration as described in the [`Camera Calibration` example](./camera-calibration.md), and verifying marker detection as outlined in the [`Marker detection`](./marker-detection.md) section, you can proceed with the hand–eye calibration process.
+
+
```
domire8:
```suggestion
```

docs/core/examples/guides/hand-eye-calibration.md
```diff
@@ -0,0 +1,63 @@
...
+This example demonstrates the eye-in-hand configuration (camera mounted on the robot arm). The procedure for the eye-to-hand configuration (static camera) follows a similar workflow.
+
```
domire8:
```suggestion
```

docs/core/examples/guides/fiducial-markers.md
MithraGhlm: Oh yeah, Marker Size is obvious. It was left empty mistakenly.

But should I omit `bundle files` from the list?

domire8: I think we could write something like
```
Bundle file: The filepath to a predefined marker bundle configuration. This additional feature is described in a separate guide (coming soon)
```

docs/core/examples/guides/fiducial-markers.md
MithraGhlm: For the part that explains obtaining the markers, I'll just put links for their download or resource github page. I think puting more detailed instructions on how to actually do this is out of this documents scope, right?

docs/core/examples/guides/hand-eye-calibration.md
MithraGhlm: Aside from adding an app, should I also expand the procedure with adding more details?

@SprGrf  @domire8

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,105 @@
...
+
+:::
+
+## Preparing fiducial markers
+
+A fiducial marker is an object placed in the field of view of an image for use as a point of reference or a measure. STag and ArUco markers are two of the common types of fiducial marker systems used for real-time 6D pose estimation. This section explains how to obtain, download and print these markers.
+
+### Obtaining markers
+
+- **ArUco marker:** ArUco markers can be generate online (e.g., from [here](https://calib.io/pages/camera-calibration-pattern-generator)), which permits choosing the dictionary, marker ID, and marker size. It can be then exported as PDF or SVG for printing.
```
domire8: I would prefer this link here: https://chev.me/arucogen/

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,105 @@
...
+
+- **STag marker:** STag marker set should be obtained from the [STag project repository](https://github.com/usrl-uofsc/stag_ros) or the generator/reference files linked by the project. The [STag project repository for ROS](https://github.com/usrl-uofsc/stag_ros) is the other place to look for marker-generation assets. In practice, you’ll want to obtain the marker PDF/SVG or generate the markers from the project’s reference generator, then print them at true size.
```
domire8: https://drive.google.com/drive/folders/0ByNTNYCAhWbIV1RqdU9vRnd2Vnc?resourcekey=0-9ipvecbezW8EWUva5GBQTQ

Public google drive where the markers can be downloaded

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Printing markres
+
+After choosing the marker family, selecting the library/dictionary, and the marker ID, download it as PDF or SVG. Use the Actual size of the marker (100% scale) for printing, so the black border and marker geometry are not resized. Also it is recommended to print with high contrast and avoid compression artifacts.
+
+If possible, print on a rigid and flat sheet of paper to reduce warping, since fiducial detection is sensitive to distortion. As another solution, you can fix the printed marker on a rigid surface, such as a piece of wood or carboard.
+
+After printing, measure the marker’s outer dimensions and compare them with the intended size from the generator. This matters because calibration will be wrong if the marker size in the software does not match the physical print.
+
+This guide provides an example of STag marker detection. Using the ArUco marker follows a very similar process.
```
domire8: This sentence now seems to be a bit in the wrong place. Maybe up top would make more sense.

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,105 @@
...
+The Camera Streamer parameters are explained in the [CameraStreamer component guide](./camera-streamer.md).
+
+:::
+
+Let's go through the parameters of the STag Detector component:
+
+- **Rate**: The rate parameter doesn't affect the behavior of the component as the detection process
+  occurs on reception of a new image.
+- **Bundle file**: The filepath to a predefined marker bundle configuration. This additional feature is described in a separate guide (coming soon).
+- **Marker selection**: The name(s) of the marker(s) that we intend to be recognized. If any of these markers enters the camera frame, the `is_any_selected_marker_detected` predicate is set to **True**. This name should always be prepended with the value of `Prefix`.
```
domire8: This is still unclear IMO. Maybe it's worth making a separate section where the two predicates are explained in detail and where you can then elaborate how the value for this parameter influences the predicate.

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,105 @@
...
+- **Marker size**: Determins the side length of a square that specifies the marker in meters.
+- **Library**: This is the ID number of the HD library utilized by STag markers. The allowed numbers are `[11, 13, 15, 17, 19, 21, 23]`.
+- **Error correction**:
```
domire8: This is still missing?

docs/core/examples/guides/hand-eye-calibration.md
```diff
@@ -0,0 +1,254 @@
+---
+sidebar_position: 15
+title: Hand-Eye calibration
+---
+
+// Should the name of the document change to "robot calibration" ?
+
+import RobotCalibrationConfiguration from './assets/robot-calibration-configuration.png'
+import RobotCameraCalibration from './assets/robot-camera-calibration-component.png'
```
domire8: this file is missing in the PR

docs/core/examples/guides/hand-eye-calibration.md
```diff
@@ -0,0 +1,254 @@
...
+<div class="text--center">
+  <img src={RobotCameraCalibration} alt="Robot Camera Calibration parameters" style={{ borderRadius: "8px" }}/>
+</div>
+
+The parameters of the `Robot Camera Calibration` component are defined as follows:
+
+- **Rate**: Determines the frequency at which transformations are acquired. This parameter does not affect the component’s behavior.
+- **Bundle file:** The filepath to a predefined marker bundle configuration. This additional feature is described in a separate guide (coming soon).
+- **Camera frame**: The name of the camera frame used in the application. This can be retrieved from the list of frames in RViz.
+- **Marker frame**: The name of the marker detected by the camera. It is indicated in the STag or Aruco marker deterctor component used in the application.
```
SprGrf:
```suggestion
- **Marker frame**: The name of the marker detected by the camera. It is indicated in the STag or Aruco marker detector component used in the application.

File: docs/core/examples/guides/hand-eye-calibration.md [hunk 40]
```diff
@@ -0,0 +1,254 @@
...
+- **Robot base frame**: The name of the robot's base frame, whichc can be found in RViz.
```
SprGrf:
```suggestion
- **Robot base frame**: The name of the robot's base frame, which can be found in RViz.

File: docs/core/examples/guides/hand-eye-calibration.md [hunk 41]
```diff
@@ -0,0 +1,254 @@
...
+        - x: 260
+          y: 60
+        - x: 260
+          y: -60
+```
+
+</details>
+
+:::tip
+Don't forget to modify the `robot_ip` acording to the ip of the robot that you are using.
```
SprGrf:
```suggestion
Don't forget to modify the `robot_ip` according to the ip of the robot that you are using.

File: docs/core/examples/guides/hand-eye-calibration.md [hunk 42]
```diff
@@ -0,0 +1,254 @@
...
+
+::: info
+
+The robot TCP can be moved by jogging or Freedrive mode using the robot's teach pendant. In the case of using a Universal Robot, AICA Studio offers the option of `Hand Guiding Controller` which facilitates and accelerates the process. This controller is described in the [`Hand Guiding Controller`](./ur-harware-interface.md) page.
+
+:::
+
+::: tip
+
+The Camera and the marker detector components might differ based on the type of hardwares being used.
```
SprGrf:
```suggestion
The Camera and the marker detector components might differ based on the type of hardware being used.

File: docs/core/examples/guides/hand-eye-calibration.md [hunk 43]
```diff
@@ -0,0 +1,254 @@
...
+
+The parameters of the `Robot Camera Calibration` component are defined as follows:
+
+- **Rate**: Determines the frequency at which transformations are acquired. This parameter does not affect the component’s behavior.
+- **Bundle file:** The filepath to a predefined marker bundle configuration. This additional feature is described in a separate guide (coming soon).
+- **Camera frame**: The name of the camera frame used in the application. This can be retrieved from the list of frames in RViz.
+- **Marker frame**: The name of the marker detected by the camera. It is indicated in the STag or Aruco marker deterctor component used in the application.
+- **Robot base frame**: The name of the robot's base frame, whichc can be found in RViz.
+- **Robot end-effector frame**: The name of the robot’s end-effector frame. This is also available in the RViz frame list.
+- **Number of recorded points**: The number of transformations from `camera` to `robot end-effector` camera to robot end-effector that are recorded.These are computed from multiple pairs of `robot end-effector` -> `robot base` and `camera` -> `marker` transformations.
```
SprGrf:
```suggestion
- **Number of recorded points**: The number of transformations from `camera` to `robot end-effector` camera to robot end-effector that are recorded. These are computed from multiple pairs of `robot end-effector` -> `robot base` and `camera` -> `marker` transformations.

File: docs/core/examples/guides/hand-eye-calibration.md [hunk 44]
```diff
@@ -0,0 +1,254 @@
...
+- **Distance between points**: The minimum spatial difference required between two consecutive transformations for them to be recorded.
+- **Epsilon time**:
+- **Calibration folder path**: The directory where the final calibration file will be stored.
+- **Calibration file**: The name of the output file generated after calibration, containing the computed calibration data.
+- **Points dataset file**: A file containing previously recorded transformations of the robot end-effector or marker, which can be reused if available.
+
+```info
+Set the `Is camera attached` check to true, if a camera is attached to the robot end-effector frame.
+```
```
SprGrf: Same in other places as well. This is also a parameter that you have not described in the list above. Maybe I would remove it from the info box to there. 
```suggestion
:::info
Set the `Is camera attached` parameter to true, if a camera is attached to the robot end-effector frame.
:::

File: docs/core/examples/guides/hand-eye-calibration.md (hunk 16 above)
Comment (SprGrf) [reply docs/core/examples/guides/hand-eye-calibration.md]: I think the level of detail is ok, but I would add a few lines on what this app is supposed to do, that for this example we used a UR5 with the hand guiding controller (link to our guide for that) and that this specific app is provided below. 

EDIT: I see that you mention that below in the info box. Maybe that fits better here.

File: docs/core/examples/guides/hand-eye-calibration.md [hunk 45]
```diff
@@ -0,0 +1,254 @@
...
+      path:
+        - x: 260
+          y: 60
+        - x: 260
+          y: -60
+```
+
+</details>
+
+:::tip
```
SprGrf: General comment on this part: the three consecutive tip/info blocks are not very readable. Maybe you could move some of this information above. 

Image

docs/core/examples/guides/hand-eye-calibration.md
```diff
@@ -0,0 +1,254 @@
...
+
+The Camera and the marker detector components might differ based on the type of hardwares being used.
+
+:::
+
+There are several ways to use the transformation information obtained in the calibration file:
+
+1. Adding the camera link to the URDF.
+2. Publish the transformation with a FrameBroadcaster manually.
+3. Using the calibration file path directly as a `Frame Broadcaster` component parameter to publish the transformations.
```
SprGrf: I think we could show how to do that here, I am not sure it's described elsewhere.

docs/core/examples/guides/fiducial-markers.md
MithraGhlm: Since this link was already in the page I mentioned in the guide, I didn't provide it exclusively. But sure, it's better this way.


Follow-up question: is it better to delete the links for the **STag project repository**  and the **STag project repository for ROS** ? Or keep them and only add the **google drive** link?

docs/core/examples/guides/hand-eye-calibration.md
MithraGhlm: I didn't  describe it since I thought it's pretty clear what it means! But maybe they should be described for the purpose of harmony and continuity.

docs/core/examples/guides/fiducial-markers.md
MithraGhlm: The problem is that there are no upper and lower limits defined for it. If the number is out of an (unspecified) range, the component doesn't work. 
I left it out on purpose to get to this issue. (But forgot to put a comment for it. Sorry)

docs/core/examples/guides/fiducial-markers.md
MithraGhlm: You mean having a separate section for explaining the 3 predicates of the STag Detector component?

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,107 @@
...
+
+A fiducial marker is an object placed in the field of view of an image for use as a point of reference or a measure. STag and ArUco markers are two of the common types of fiducial marker systems used for real-time 6D pose estimation. This section explains how to obtain, download and print these markers.
+
+### Obtaining markers
+
+- **ArUco marker**: ArUco markers can be generate online (e.g., from [here](https://chev.me/arucogen/)), which permits choosing the dictionary, marker ID, and marker size. It can be then exported as PDF or SVG for printing.
+
+- **STag marker**: STag marker set can be either downloaded from [google public drive](https://drive.google.com/drive/folders/0ByNTNYCAhWbIV1RqdU9vRnd2Vnc?resourcekey=0-9ipvecbezW8EWUva5GBQTQ) or obtained from the [STag project repository](https://github.com/usrl-uofsc/stag_ros) or the generator/reference files linked by the project. The [STag project repository for ROS](https://github.com/usrl-uofsc/stag_ros) is the other place to look for marker-generation assets. In practice, you’ll want to obtain the marker PDF/SVG or generate the markers from the project’s reference generator, then print them at true size.
```
domire8: Link 2 and 3 are the same. Also, if you want to link to that repo, please do so with the ros2-devel branch: https://github.com/usrl-uofsc/stag_ros/tree/ros2-devel

domire8 (commented):
Could you also apply the formatting by the "prettier" extension? It's possible that you also have to adjust the print width setting for the extension to be 120 

Image

docs/core/examples/guides/fiducial-markers.md
domire8: Ah I didn't see that the repo links the drive as well. I think the repo can still be linked, but also see my comment below because the two links go to the same repository. So essentially we probably end up with two links, the drive and the repo

docs/core/examples/guides/fiducial-markers.md
domire8: Yes

docs/core/examples/guides/fiducial-markers.md
domire8: It would be good to figure out what the limits are then. Like that we can also provide more insight what that parameter actually does.

MithraGhlm: The lower limit in the code is 0, and the upper bound comes from the upstream STag library, in which the rule is:
`0 ≤ errorCorrection ≤ (libraryHD - 1) / 2`

With the default library value of 15, the maximum valid error correction is 7.

There’s one mismatch though.  In the file
`aica-technology/dynamic components/source/core_vision_components/extension_descriptions/core_vision_components_pose_detection_stag_bundle_registration.yaml`

and the file:

`aica-technology/dynamic-components/source/core_vision_components/extension_descriptions/core_vision_components_pose_detection_stag_detector.yaml`


 a minimum of 1 is indicated, while the code accepts 0.

```
- display_name: Error correction
parameter_name: error_correction
description: The error correction tolerance. Lower values are strict, higher values are more permissive
default_value: 7
parameter_type: int
validation:
range:
minimum: 1
```

MithraGhlm (commented):
Upper and Lower limits for Error_correction of STag Detector

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Obtaining markers
+
+- **ArUco marker**: ArUco markers can be generate online (e.g., from [here](https://chev.me/arucogen/)), which permits choosing the dictionary, marker ID, and marker size. It can be then exported as PDF or SVG for printing.
+
+- **STag marker**: STag marker set can be either downloaded from [google public drive](https://drive.google.com/drive/folders/0ByNTNYCAhWbIV1RqdU9vRnd2Vnc?resourcekey=0-9ipvecbezW8EWUva5GBQTQ) or obtained from the [ROS2 STag project repository](https://github.com/usrl-uofsc/stag_ros/tree/ros2-devel) or the generator/reference files linked by the project. In practice, you’ll want to obtain the marker PDF/SVG or generate the markers from the project’s reference generator, then print them at true size.
+
+### Printing markers
+
+After choosing the marker family, selecting the library/dictionary, and the marker ID, download it as PDF or SVG. Use the Actual size of the marker (100% scale) for printing, so the black border and marker geometry are not resized. Also it is recommended to print with high contrast and avoid compression artifacts.
```
domire8:
```suggestion
After choosing the marker family, selecting the library/dictionary, and the marker ID, download it as PDF or SVG. Use the actual size of the marker (100% scale) for printing, so the black border and marker geometry are not resized. Also it is recommended to print with high contrast and avoid compression artifacts.
```

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,121 @@
...
+
+If possible, print on a rigid and flat sheet of paper to reduce warping, since fiducial detection is sensitive to distortion. As another solution, you can fix the printed marker on a rigid surface, such as a piece of wood or cardboard.
+
+After printing, measure the marker’s outer dimensions and compare them with the intended size from the generator. This matters because calibration will be wrong if the marker size in the software does not match the physical print.
```
domire8: The second phrase sounds a bit like the print absolutely has to be perfect in order for the rest to work. It's subtle but maybe I would go for something like this:

After printing, measure the marker's outer dimensions since those might still be different from the desired ones. The true physical size will be required below to ensure that the calibration is precise.

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,121 @@
...
+</div>
+
+- **Is a marker bundle detected**: If a registered group of markers is detected by the camera, this parameter will be set to **True**. Otherwise it will remain **False**.
+
+## STag Detector parameters
+
+- **Rate**: The rate parameter doesn't affect the behavior of the component as the detection process
+  occurs on reception of a new image.
+- **Bundle file**: The filepath to a predefined marker bundle configuration. This additional feature is described in a separate guide (coming soon).
+- **Marker selection**: The name(s) of the marker(s) that we intend to be recognized. If any of these markers enters the camera frame, the `is_any_selected_marker_detected` predicate is set to **True**. This name should always be prepended with the value of `Prefix`.
```
domire8: "we intend to be recognized" is wrong

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,121 @@
...
+- **Marker size**: Determines the side length of a square that specifies the marker in meters.
+- **Library**: This is the ID number of the HD library utilized by STag markers. The allowed numbers are `[11, 13, 15, 17, 19, 21, 23]`.
+- **Error correction**:
+- **Prefix**: This prefix is used for marker names.
+
+:::tip
+
+If a decision needs to be made based on the existence of a specific STag marker in the camera frame, its name should be indicated in the `Marker Selection` parameter.
```
domire8: This tip feels a bit off here, if anything it could be explained in the predicates section.

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,121 @@
...
+
+:::
+
+:::info
+
+If any marker is detected in the camera frame at all, the `is_any_marker_detected` predicate is set to **True**.
```
domire8: This is now explained in the predicates section no?

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,121 @@
...
+
+<div class="text--center">
+  <img src={stagDetectorExample} alt="CameraStreamer configuration alongside STagDetector component" style={{ borderRadius: "8px" }}/>
+</div>
+
+:::info
+The Camera Streamer parameters are explained in the [CameraStreamer component guide](./camera-streamer.md).
+:::
+
+## STag Detector predicates
```
domire8, line 101: You asked me about this and I couldn't give a good answer then but now I think the parameters section has to come first

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Preparing fiducial markers
+
+A fiducial marker is an object placed in the field of view of an image for use as a point of reference or a measure. STag and ArUco markers are two of the common types of fiducial marker systems used for real-time 6D pose estimation. This section explains how to obtain, download and print these markers.
+
+### Obtaining markers
+
+- **ArUco marker**: ArUco markers can be generate online (e.g., from [here](https://chev.me/arucogen/)), which permits choosing the dictionary, marker ID, and marker size. It can be then exported as PDF or SVG for printing.
+
+- **STag marker**: STag marker set can be either downloaded from [google public drive](https://drive.google.com/drive/folders/0ByNTNYCAhWbIV1RqdU9vRnd2Vnc?resourcekey=0-9ipvecbezW8EWUva5GBQTQ) or obtained from the [ROS2 STag project repository](https://github.com/usrl-uofsc/stag_ros/tree/ros2-devel) or the generator/reference files linked by the project. In practice, you’ll want to obtain the marker PDF/SVG or generate the markers from the project’s reference generator, then print them at true size.
```
domire8:
```suggestion
- **STag marker**: STag marker set can be either downloaded from [public Google Drive](https://drive.google.com/drive/folders/0ByNTNYCAhWbIV1RqdU9vRnd2Vnc?resourcekey=0-9ipvecbezW8EWUva5GBQTQ) or obtained from the [ROS2 STag project repository](https://github.com/usrl-uofsc/stag_ros/tree/ros2-devel) or the generator/reference files linked by the project. In practice, you’ll want to obtain the marker PDF/SVG or generate the markers from the project’s reference generator, then print them at true size.
```

docs/core/examples/guides/fiducial-markers.md
domire8: `0 ≤ errorCorrection ≤ (libraryHD - 1) / 2` ah that is a shame that the error bounds depend on the library selection.. This is something we can't validate easily in Studio directly.

You could then open a PR on the relevant description files to set the minimum to 0 and also to set the maximum to `(23 - 1) / 2 = 11`. You could even improve the component implementation such that a detail ROS log is spit out when the error correction is too high for the given library.

docs/core/examples/guides/fiducial-markers.md
MithraGhlm: true

docs/core/examples/guides/fiducial-markers.md
MithraGhlm: I displaced it to the parameter section, where I explained the `Marker selection` parameter. I feel it matches there better. Do you agree?

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,109 @@
...
+- **Is any selected marker detected**: If one or more marker names are indicated in the `Marker selection` parameter, and if any of them appears in the camera frame, this predicate with be set to **True**. If the names of none of the markers present in the camera frame is indicated as a parameter, this predicate will remain **False**.
+  In the screenshot below the name of the marker appearing in the camera frame matches the name indicated in the `Marker selection` parameter.
+
+<div class="text--center">
+  <img src={stagMarkerNumZero} alt="Is any of the selected markers detected" style={{ borderRadius: "8px" }}/>
+</div>
+
+- **Is a marker bundle detected**: If a registered group of markers is detected by the camera, this parameter will be set to **True**. Otherwise it will remain **False**.
+
+After setting up the proper parameters for Camera Streamer and STag Detector:
```
domire8, line 127: I think this could be in it's own section "Running the application" or something like that and we should attach the YAML content too

domire8 (commented):
Nice, this is really coming together nicely. As we enter the final stages, could you now please try to format the file fiducial-markers.md?

docs/core/examples/guides/fiducial-markers.md
MithraGhlm: Is there a need to investigate ArUco as well?

MithraGhlm:
> Nice, this is really coming together nicely. As we enter the final stages, could you now please try to format the file fiducial-markers.md?

I am using Prettier. I also changed the print width to 120.  Isn't it working as intended?

docs/core/examples/guides/fiducial-markers.md
MithraGhlm: true

domire8:
> I am using Prettier. I also changed the print width to 120. Isn't it working as intended?

Well no, most lines are still way longer than 120 characters

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,205 @@
...
+:::info
+The Camera Streamer parameters are explained in the [CameraStreamer component guide](./camera-streamer.md).
+:::
+
+## STag Detector parameters
+
+- **Rate**: The rate parameter doesn't affect the behavior of the component as the detection process
+  occurs on reception of a new image.
+- **Bundle file**: The filepath to a predefined marker bundle configuration. This additional feature is described in a separate guide (coming soon).
+- **Marker selection**: The name(s) of the marker(s) that we want to recognize. If any of these markers enters the camera frame, the `is_any_selected_marker_detected` predicate is set to **True**. Also if a decision needs to be made based on the existence of a specific STag marker in the camera frame, its name should be indicated in this parameter. The markers name should always be prepended with the value of `Prefix`.
```
domire8:
```suggestion
- **Marker selection**: The name(s) of the marker(s) that we want to recognize. If any of these markers enters the camera frame, the `is_any_selected_marker_detected` predicate will switch to **True** (see more in the section below). The marker names should always be prepended with the value of `Prefix`.
```

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,205 @@
...
+- **Marker size**: Determines the side length of a square that specifies the marker in meters.
```
domire8, line 90:
```suggestion
- **Marker size**: The measured side length of the marker in meters.
```

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,205 @@
...
+- **Library**: This is the ID number of the HD library utilized by STag markers. The allowed numbers are `[11, 13, 15, 17, 19, 21, 23]`.
```
domire8:
```suggestion
- **Library**: This is the ID number of the HD library utilized by STag markers. Allowed options are `[11, 13, 15, 17, 19, 21, 23]`.
```

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,205 @@
...
+- **Error correction**: This parameter sets how permissive the detector is. Meaning it controls how many bit errors the STag detector is allowed to tolerate when decoding a marker ID. Lower values mean stricter matching, but detection would be less robust to blur, noise, bad lighting, or partial image degradation. Higher values mean more tolerant matching. The detector can still recognize a marker even if some bits are read incorrectly, so detection is more robust, but the risk of wrong matches increases. The lower and upper limits for this parameter are 0 and 11 respectively (inclusive).
```
domire8:
```suggestion
- **Error correction**: This parameter sets how permissive the detector is. It controls how many bit errors the STag detector is allowed to tolerate when decoding a marker ID. Lower values mean stricter matching, but detection would be less robust to blur, noise, bad lighting, or partial image degradation. Higher values mean more tolerant matching. The detector can still recognize a marker even if some bits are read incorrectly, so detection is more robust, but the risk of wrong matches increases. The lower and upper limits for this parameter are 0 and 11 respectively.
```

domire8 (commented):
I have a few last nitpicks for the fiducial markers guide.

In order to proceed here, please
1. revert the changes from the ur-hardware-interface.md file, you can do that in #485 
2. Remove the hand-eye-calibration guide from this PR. We'll make a separate PR for it.
3. Format the fiducial-markers.md file such that no lines are longer than 120 characters

docs/core/examples/guides/fiducial-markers.md
domire8: As far as I can see, Aruco doesn't have an error correction parameter, so I think there is no need. Double check please.

MithraGhlm:
> > I am using Prettier. I also changed the print width to 120. Isn't it working as intended?
> 
> Well no, most lines are still way longer than 120 characters

I added `"prettier.proseWrap": "always"`,
seems like without this, Prettier often preserves existing line breaks and may ignore `printWidth`.

MithraGhlm:
> In order to proceed here, please
> 
> 1. revert the changes from the ur-hardware-interface.md file, you can do that in #485

Why should they be reverted? 
The changes are not big, just some tiny modifications, such as Bolding of some words, made to make the formatting look more in sync across different pages. Does it cause any problem?

domire8:
> The changes are not big, just some tiny modifications, such as Bolding of some words, made to make the formatting look more in sync across different pages. Does it cause any problem?

Unless we have actual guidelines how we want to improve style and formatting, there is no need increasing the scope of this PR to files that are unrelated to hand-eye calibration. It's good practice to keep scopes of PRs narrow.

I'm not against the changes alltogether, but this is not the place to do it. A reformatting PR would be more appropriate

MithraGhlm:
> In order to proceed here, please

> 2. Remove the hand-eye-calibration guide from this PR. We'll make a separate PR for it.

Before I change anything there, do you want the existing review comments be preserved or is it acceptable to move it to a new branch and create a fresh PR?

This is what I want to do: (Since the name of  this PR is already `HE calibration`, I'll move the `Fiducial Markers` to a new PR)
1. Make a new branch for `Fiducial Markers`
2. Delete the `fiducial-markers.md ` file and all media related to it

So the existing PR would be only for `HE calibration`.

Then I'll open a new PR for the `Fiducial Markers` and rebase on `main`.

Is this workflow accepted?

MithraGhlm:
> Unless we have actual guidelines how we want to improve style and formatting, there is no need increasing the scope of this PR to files that are unrelated to hand-eye calibration. It's good practice to keep scopes of PRs narrow.
> 
> I'm not against the changes alltogether, but this is not the place to do it. A reformatting PR would be more appropriate

As you correctly commented once on the #485 (_This PR contains all commits from https://github.com/aica-technology/api/pull/482, could you untangle this?_), I untangled #482 from #485. 
The initial problem was due to me forgetting to switch to `main` before making the new branch. But then I rebased it on the `main` and the problem was resolved. Now #482 and #485 are 2 different PRs. 

So I don't understand what you mean by: _increasing the scope of this PR to files that are unrelated to hand-eye calibration. It's good practice to keep scopes of PRs narrow._

Since we don't have any specific guidelines yet, how about  just merging the small changes and closing #485, instead of reverting it?

domire8:
All 4 of my previous review comments are still unresolved

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,238 @@
...
+  markers present in the camera frame is indicated as a parameter, this predicate will remain **False**. In the
+  screenshot below the name of the marker appearing in the camera frame matches the name indicated in the
+  `Marker selection` parameter.
+
+<div class="text--center">
+  <img src={stagMarkerNumZero} alt="Is any of the selected markers detected" style={{ borderRadius: "8px" }}/>
+</div>
+
+- **Is a marker bundle detected**: If a registered group of markers is detected by the camera, this parameter will be
+  set to **True**. Otherwise it will remain **False**.
```
domire8, line 123:
```suggestion
- **Is any marker detected**: This predicate will be set to **True** if any marker is detected in the camera frame, even
  though its name is not indicated in the 'Marker Selection' parameter. As you can see in the screenshot below, the
  marker name specified in the `Marker selection` parameter is stag_1, but the marker recognized in the camera frame is
  stag_0, yet `Is any marker detected` predicate is set to **True**.
  <div class="text--center">
    <img src={stagMarkerNumOne} alt="Is any marker detected at all" style={{ borderRadius: "8px" }}/>
  </div>
- **Is any selected marker detected**: If one or more marker names are indicated in the `Marker selection` parameter,
  and if any of them appears in the camera frame, this predicate with be set to **True**. If the names of none of the
  markers present in the camera frame is indicated as a parameter, this predicate will remain **False**. In the
  screenshot below the name of the marker appearing in the camera frame matches the name indicated in the
  `Marker selection` parameter.
  <div class="text--center">
    <img src={stagMarkerNumZero} alt="Is any of the selected markers detected" style={{ borderRadius: "8px" }}/>
  </div>
- **Is a marker bundle detected**: If a registered group of markers is detected by the camera, this parameter will be
  set to **True**. Otherwise it will remain **False**.
```
purely a formatting change here. By adding the indents before the image `<div>` elements, the images will have the same width as the enumeration. 

Image

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,238 @@
...
+3. In RViz, select _Add > By topic > /stag_detector/annotated_image > Image_. This adds a panel that shows the live
+   image. The marker should be detected in the camera.
+
+<div style={{ display: "flex", justifyContent: "center" }}>
+  <video autoPlay loop muted playsInline style={{ maxWidth: "100%", borderRadius: "8px" }}>
+    <source src={stagMarkerDetection} type="video/webm" />
+    STag marker detection video.
+  </video>
+</div>
+
```
domire8, line 140:
```suggestion
<br></br>
```
This adds a bit more whitespace after the image

domire8 (commented):
Just two last formatting nitpicks. Otherwise all good!

docs/core/examples/guides/fiducial-markers.md
MithraGhlm: So the image being aligned with the text is the preferred format over the image being aligned with the bullet points?
Great. Got it :)

MithraGhlm: Should other images be also shifted a tab to the right, or it only applies to those who are next to the bullet points?

docs/core/examples/guides/fiducial-markers.md
MithraGhlm: Cool, the little space between them was bothering me a bit, but I thought it's normal to be that way.

docs/core/examples/guides/fiducial-markers.md
domire8: I think the ones in enumerations only is better. The main point is that the images should look like they are part of the enumeration, not that they interrupt the alignment of the enumerations.

domire8 (approved):
Alright looks good now! Good work with this :+1:

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,240 @@
...
+          y: 360
+        - x: 660
+          y: 300
+```
+
+</details>
+
+:::info
+The process for using ArUco markers follows a similar process.
+:::
```
SprGrf: I would remove this part altogether, or rephrase. Repetition of "process", it's a bit much in the eye, two consecutive blue blocks, and you mention the same thing in line 23 either way.

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,240 @@
...
+A fiducial marker is an object placed in the field of view of an image for use as a point of reference or a measure.
+STag and ArUco markers are two of the common types of fiducial marker systems used for real-time 6D pose estimation.
+This section explains how to obtain, download and print these markers.
+
+### Obtaining markers
+
+- **ArUco marker**: ArUco markers can be generate online (e.g., from [here](https://chev.me/arucogen/)), which permits
+  choosing the dictionary, marker ID, and marker size. It can be then exported as PDF or SVG for printing.
+
+- **STag marker**: STag marker set can be either downloaded from
```
SprGrf:
```suggestion
- **STag marker**: STag marker sets can be either downloaded from

File: docs/core/examples/guides/fiducial-markers.md [hunk 64]
```diff
@@ -0,0 +1,240 @@
...
+  [public Google Drive](https://drive.google.com/drive/folders/0ByNTNYCAhWbIV1RqdU9vRnd2Vnc?resourcekey=0-9ipvecbezW8EWUva5GBQTQ)
+  or obtained from the [ROS2 STag project repository](https://github.com/usrl-uofsc/stag_ros/tree/ros2-devel) or the
+  generator/reference files linked by the project. In practice, you’ll want to obtain the marker PDF/SVG or generate the
+  markers from the project’s reference generator, then print them at true size.
+
+### Printing markers
+
+After choosing the marker family, selecting the library/dictionary, and the marker ID, download it as PDF or SVG. Use
```
SprGrf:
```suggestion
After choosing the marker family, selecting the library/dictionary and the marker ID, download it as PDF or SVG. Use

File: docs/core/examples/guides/fiducial-markers.md [hunk 65]
```diff
@@ -0,0 +1,240 @@
...
+:::info
+The Camera Streamer parameters are explained in the [CameraStreamer component guide](./camera-streamer.md).
+:::
+
+## STag Detector parameters
+
+- **Rate**: The rate parameter doesn't affect the behavior of the component as the detection process occurs on reception
+  of a new image.
+- **Bundle file**: The filepath to a predefined marker bundle configuration. This additional feature is described in a
+  separate guide (coming soon).
```
SprGrf: Sorry maybe I missed a part of the discussion, where are we going to describe this?

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,240 @@
...
+- **Marker selection**: The name(s) of the marker(s) that we want to recognize. If any of these markers enters the
+  camera frame, the `is_any_selected_marker_detected` predicate will switch to **True** (see more in the section below).
+  The marker names should always be prepended with the value of `Prefix`.
+- **Marker size**: The measured side length of the marker in meters.
+- **Library**: This is the ID number of the HD library utilized by STag markers. Allowed options are
+  `[11, 13, 15, 17, 19, 21, 23]`.
```
SprGrf: Nit, but there is something I don't like here. Both the "[]" and the highlight is a bit much I think. I would remove the "[]", there is no need to format this as a list.

docs/core/examples/guides/.vscode/settings.json
SprGrf, line 1: Is this intentional or a leftover?

SprGrf (commented):
Nice work, sorry it took me so long to return to this!

docs/core/examples/guides/fiducial-markers.md
domire8: either here or in a new document, I just think that now we should rather do the calibration page than spending too much time with the bundle feature because it's less common

SprGrf: Agreed, but I fear it will be forgotten quickly. I would be in favor of removing the note about it being described in a different guide and creating and issue for that.

domire8: Would you delete the entire lines 85 and 86 then or how?

MithraGhlm:
> Nice work, sorry it took me so long to return to this!

No worries :) It's nice you're back.

docs/core/examples/guides/fiducial-markers.md
MithraGhlm: I don't know, to be honest. If I'm not mistaken, this sentence was suggested by @domire8 and I applied it. I didn't ask where, unfortunately.

docs/core/examples/guides/fiducial-markers.md
MithraGhlm: You're right. I'll rather remove it.

docs/core/examples/guides/.vscode/settings.json
MithraGhlm: It was automatically added by Prettier, when I told it not to consider these words as typos!
Will it cause any problem? Should I remove it?

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,240 @@
...
+<div style={{ display: "flex", justifyContent: "center" }}>
+  <video autoPlay loop muted playsInline style={{ maxWidth: "100%", borderRadius: "8px" }}>
+    <source src={stagMarkerDetection} type="video/webm" />
+    STag marker detection video.
+  </video>
+</div>
+
+<br></br>
+
+The following YAML snippet contains the full application of the STag Detector above:
```
MithraGhlm, line 143: Is the distance between this line and the video above it enough? They look like they're too close in my eyes.

docs/core/examples/guides/fiducial-markers.md
SprGrf: Yeah, maybe i would just leave only the "The filepath to a predefined marker bundle configuration" sentence. And create an issue to remember to add to that.

docs/core/examples/guides/.vscode/settings.json
SprGrf: I don't see it as an issue, it's just that usually we don't commit such types of changes. @domire8 what do you think?

docs/core/examples/guides/fiducial-markers.md
SprGrf: I think it's fine.

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,241 @@
...
+        - x: 660
+          y: 360
+        - x: 660
+          y: 300
+```
+
+</details>
+
+
+>>>>>>> be81941 (some small final formatting changes)
```
SprGrf: Something left over here

docs/core/examples/guides/.vscode/settings.json
domire8: I don't mind it being here, even though it might be more correct to add that to the individual global vscode settings and not the project settings.

docs/core/examples/guides/.vscode/settings.json
```diff
@@ -0,0 +1 @@
+
```
SprGrf: Thanks for this! Then this file and its parent folder can be removed altogether I think.

MithraGhlm: sure.

docs/core/examples/guides/fiducial-markers.md
```diff
@@ -0,0 +1,240 @@
...
+        - x: 660
+          y: 360
+        - x: 660
+          y: 300
+```
+
+</details>
+
+
+
```
domire8, line 240:
```suggestion

```

domire8 (approved):
Yes very nice :+1:

PR #487, opened 2026-05-19
feat(docs): add proxies documentation

LouisBrunner:
## Description

@domire8 

## Review guidelines

## Checklist before merging:


eeberhard (approved):
well explained and written, good from my side. I think we can already release these docs prior to Launcher v2.1 release, would you agree @rrosa24?

domire8 (commented):
~I'm a bit hesitant to put a docker desktop screenshot here because for Linux users we discourage the use of docker desktop. Let me investigate what we could put instead.~

Sorry I will review it properly, I think it's very well written and I was a bit too fast judging it from the screenshot

rrosa24 (approved):
Yes looks good thank you, would be fine putting out pre- v2.1

PR #491, opened 2026-06-10
Docs/hand eye calibration guide

MithraGhlm:
The  documentation for camera frame calibration guide.
This PR is dedicated to calibrating the robot.


SprGrf:
We can proceed with this one now, but I think it includes commits from the last merged PR. Could you please remove the fiducial related stuff?

MithraGhlm:
> We can proceed with this one now, but I think it includes commits from the last merged PR. Could you please remove the fiducial related stuff?

Sure. So what I intended to do was to have the `HE calibration` branched from the `Fiducial Marker` branch, but if I understand now correctly, you find it better to be as an independent branch, based on `main`. Right?

SprGrf:
Since main now has the fiducial markers merged , I think this would work yes.

